PR #45, opened 2021-10-04
Hamzahsn/feat/show notification when not secure

hamzahsn:
This PR...

## Changes

- This PR will show a nice notification when the host protocol is not secure.

## Fixes

-

## How to test it

-

## screenshots

- 
Screenshot 2021-10-06 at 14 29 59

## Checklist

- [x ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


hamzahsn:
Need to confirm the styles and the text and then we can run it after the Get tests modal.

olensmar:
it's not enough to show a notification - we need to tell users what to do about it... we'll need to create a page in the documentation that we can link to from here...

hamzahsn:
So you suggest something like: `Dashboard is using non-secure protocol! find more here` ? 🤔

hamzahsn:
Needs to update the link to the docs in the error notification as write now it's in progress.

src/App.tsx
```diff
@@ -110,7 +110,7 @@ function App() {
 
     if (!isHostProtocolSecure()) {
       showSmallError(`Dashboard is using non-secure protocol! Read more
```
olensmar, line 112: remove "Read more" on this line - not necessary to have it both with and without link

src/App.tsx
```diff
@@ -110,7 +110,7 @@ function App() {
...
-      <a href='https://github.com/kubeshop/kubtest/blob/main/docs/getting-started.md' target="_blank" rel="noopener">Read more</a>`);
+      <a href='https://kubeshop.github.io/kubtest/installing/' target="_blank" rel="noopener">Read more</a>`);
```
olensmar, line 113: we will  update the link before the release - this is ok for now

src/App.tsx
hamzahsn: ooops missed that

src/App.tsx
olensmar: yes - we'll update when we have the content available

olensmar (approved):

PR #56, opened 2021-10-06
Hamzahsn/feat/show modal only when api does not endpoint exists

hamzahsn:
This PR...

## Changes

- This PR is about rendering the modal where user specify the api endpoint if no API were specified while opening the dashboard or no API found in localStorage.

## Fixes

-

## How to test it

-

## screenshots

- 

## Checklist

- [x ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/atoms/Modal/Modal.tsx
+++ b/src/components/atoms/Modal/Modal.tsx
@@ -1,0 +1,101 @@
+import React from 'react';
+import styled from 'styled-components';
+import {Modal} from 'antd';
+
+import {Button, LabelInput, Typography} from '@atoms';
+import {
+  validateUrl,
+  matchEndpointProtocolWithHostProtocol,
+  removeDuplicatesInQueryString,
+  checkApiEndpointProtocol,
+} from '@utils/validate';
+
+const StyledSearchUrlForm = styled.form`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+`;
+
+const modalStyles = {
+  backgroundColor: 'var(--color-dark-primary)',
+  height: '300px',
+  overflow: 'hidden',
+};
+
+const StyledFormContainer = styled.div`
+  display: flex;
+  align-items: center;
+`;
+
+interface IUrlEndpoint {
+  apiEndpoint: string;
+}
+
+interface IModal {
+  isModalVisible: (isVisible: boolean) => void;
+  visible?: boolean;
+}
+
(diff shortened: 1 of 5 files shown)


hamzahsn:
When using this feature with staged env on Netlify - API does not work - CORS problem, needs to check that with @fog1985 and @nicufk

src/components/atoms/Modal/Modal.tsx
```diff
@@ -0,0 +1,101 @@
...
+        </Typography>
+        <StyledFormContainer>
+          <LabelInput
+            id="url"
+            name="url"
+            onChange={(event: React.ChangeEvent<HTMLInputElement>) => handleChangeApiEndpoint(event, 'apiEndpoint')}
+            defaultValue={apiEndpoint.apiEndpoint}
+          />
+          <Button type="submit" disabled={!validUrl} disableFilter>
+            Get tests
```
olensmar, line 93: I think we said "Get Results" for this button?

src/components/atoms/Modal/Modal.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <Typography variant="secondary">
+          The endpoint needs to be accessible from your browser and will be used to retrieve test results only.
+        </Typography>
+        <StyledFormContainer>
+          <LabelInput
+            id="url"
+            name="url"
+            onChange={(event: React.ChangeEvent<HTMLInputElement>) => handleChangeApiEndpoint(event, 'apiEndpoint')}
+            defaultValue={apiEndpoint.apiEndpoint}
+          />
```
olensmar, line 91: I noticed in the current demo dashboard that selecting an endpoint provided by the browser-history does not enable the button, you have to make an actual change with the keyboard  - is this still an issue here?

src/components/atoms/Modal/Modal.tsx
hamzahsn: You're right, but I have updated this one in another PR which I'm almost done from it https://github.com/kubeshop/kubtest-dashboard/pull/44

src/components/atoms/Modal/Modal.tsx
hamzahsn: this one I have also tried to handle it in the same PR which mentioned above.

olensmar (approved):
ok for now - let's see how it all comes together!

PR #77, opened 2021-10-12
Hamzahsn/feat/get today tests instead latest

hamzahsn:
This PR...

## Changes

- Allow user to grab tests for `today` instead of `latest` as it was not clear.

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/utils/testFilters.ts
+++ b/src/utils/testFilters.ts
@@ -1,4 +1,4 @@
-import { getDate, getLatestDate } from "./formatDate";
+import { getDate, getTodayTests } from "./formatDate";
 
 export const filterTestsExecution = (tests: any, _filters: any) => {
     let filteredTestsExecution = tests;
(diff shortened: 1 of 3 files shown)


src/utils/testFilters.ts
```diff
@@ -1,30 +1,65 @@
...
 
 export const filterTestsExecution = (tests: any, _filters: any) => {
-    let filteredTestsExecution = tests;
+  let filteredTestsExecution = tests;
 
-    if ((_filters?.filter?.includes('error') || _filters?.filter?.includes('success') || _filters?.filter?.includes('pending'))) {
-        const filteredTests = filteredTestsExecution?.results?.filter((test: any) => _filters?.filter?.includes(test.status));
-        filteredTestsExecution = { ...filteredTestsExecution, results: filteredTests };
-    }
+  if (
```
olensmar, line 6: filtering should be entirely done server-side - since the response is paged you might not have any results to show if you filter on the client

PR #79, opened 2021-10-13
Hamzahsn/fix/update styles

hamzahsn:
This PR...

## Changes

- Updating styles depending on the design.

## Fixes

- The table should go all the way to the bottom of the page
- Table’s borders to the right should be visible, actually all tables should have all borders visible
- all the heights marked with an arrow in the dashboard are too high compared to the design
- the design uses gray font for labels/texts, the dashboard uses white/bold ?
- spacing between step rows is too large
- the filter buttons are too large and spaced too far apart compared to the design
- font used in step labels is not same as in design
- when selecting an execution, the selected item should be marked as selected (maybe highlighted or similar). Idea is that the users know which item is being shown in the detailed view

## How to test it

-

## screenshots

-

## Checklist

- [x ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


hamzahsn:
Need to test this with other team members screens cause dev design is aligned(sizes) but prod design is not aligned and it is with the latest changes with `main` branch, maybe something wrong with Netlify cache?!  Even though I build the project and served it locally and still, design(sizes) is aligned already!

Screenshot 2021-10-13 at 02 50 25

PR #137, opened 2021-11-22
Update page size on initial load to 10 per page

hamzahsn:
This PR...

## Changes

- On page load we currently load 100 results per page which could be not a good way for user to see the results, so we changed that to 10 results per page.

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


fog1985 (approved):
LGTM

PR #504, opened 2023-01-06
execute clipboard.writeText, if isSecureContext

chooco13:
in `tests` - `setting` - `Definition`
if you click code when insecure network (http),  the screen turn white with error in console
because of trying to write in clipboard in insecure network (http)

```
TypeError: Cannot read properties of undefined (reading 'writeText')
    at useCopyToClipboard.ts:22:27
    at eu (react-dom.production.min.js:242:332)
    at wl (react-dom.production.min.js:285:111)
    at react-dom.production.min.js:282:358
    at Ol (react-dom.production.min.js:280:398)
    at cl (react-dom.production.min.js:272:439)
    at Ho (react-dom.production.min.js:127:105)
    at react-dom.production.min.js:266:269
    
react-dom.production.min.js:127 Uncaught TypeError: Cannot read properties of undefined (reading 'writeText')
    at useCopyToClipboard.ts:22:27
    at eu (react-dom.production.min.js:242:332)
    at wl (react-dom.production.min.js:285:111)
    at react-dom.production.min.js:282:358
    at Ol (react-dom.production.min.js:280:398)
    at cl (react-dom.production.min.js:272:439)
    at Ho (react-dom.production.min.js:127:105)
    at react-dom.production.min.js:266:269
```

## Changes & Fixes
- check the network is secure(https) or insecure network (http), and skip if insecure

## How to test it
- if you click code when insecure network (http),  the screen not turn to white

### DIFF
--- a/src/hooks/useCopyToClipboard.ts
+++ b/src/hooks/useCopyToClipboard.ts
@@ -14,7 +14,9 @@ export const useCopyToClipboard = (textToCopy: string, options: CopyToClipboardO
     let timeout: NodeJS.Timeout;
 
     if (isCopied) {
-      navigator.clipboard.writeText(textToCopy);
+      if (window.isSecureContext) {
+        navigator.clipboard.writeText(textToCopy);
+      }
       timeout = setTimeout(() => {
         setCopyToClipboardState(false);
       }, timeoutInMs);


pavloburchak:
Hello, @chooco13! First of all, thank you for contribution, we really appreciate this!

It is good that you found this bug, but the solution doesn't really solves the problem. When accessing by insecure network, user still won't be able to copy needed content.

I can see a better solution to this issue. We can remove the `Copy` button at all for insecure network and replace it with `Download` button. In this case Yaml file will be generated and downloaded on click. 

Will it work for you? What do you think about this?

chooco13:
Thank you for your answer @pavloburchak 

That seems like a good way. : )
If there are users who need to copy, they can try copying by dragging.

pavloburchak:
Alright!

So I am going to close this PR and make this improvement in next release.
Here is an issue https://github.com/kubeshop/testkube/issues/3022

PR #505, opened 2023-01-09
Remove defer from env-config load

kamilkrampa:
Hey,
with defer `REACT_APP_API_SERVER_ENDPOINT` was always null for me. You should be able to easily reproduce it by building the image on the main branch and running it with `REACT_APP_API_SERVER_ENDPOINT` set. It'll show you modal with empty url. Removing defer fixes it for me. 

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -28,7 +28,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Testkube</title>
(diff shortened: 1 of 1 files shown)


pavloburchak:
@kamilkrampa thank you for the contribution. Approved

pavloburchak (approved):

PR #508, opened 2023-01-12
add radio button for executorType in AddExecutorsModal.tsx and change condition of customExecutors filter 

chooco13:
## Changes

I'm currently making executor for "playwright".
https://github.com/chooco13/testkube-executor-playwright

I found out that the definition value did not come out as I wanted
when creating an executor through the dashboard

The definition value I was expecting

```
spec:
   image: xxx.xxx.xxx.xxx:5000/playwright-executor
   types:
     -playwright/project
```

definition value that actually came out

```
spec:
   executor_type: container
   image: xxx.xxx.xxx.xxx:5000/playwright-executor
   types:
     -playwright/project
```

so I did

- add radio button for executorType in AddExecutorsModal.tsx
- change condition of customExecutors filter

However, even with this
it doesn't come out the way I want it to.

because of below code line
https://github.com/kubeshop/testkube/blob/main/internal/app/api/v1/executor.go#L28
So I also suggest adding ` && executor.Spec.Image == "" ` to the condition on that line.

## Fixes

-

## How to test it

- 

## screenshots

Modal image
customExecutors filter

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/custom-antd/Radio/Radio.styled.tsx
+++ b/src/components/custom-antd/Radio/Radio.styled.tsx
@@ -1,0 +1,10 @@
+import {Radio, RadioProps} from 'antd';
+
+import styled from 'styled-components';
+
+export interface ICustomRadioProps extends RadioProps {
+  checked?: boolean;
+}
+
+export const AntdCustomStyledRadio = styled(Radio)<ICustomRadioProps>`
+`;
--- a/src/components/custom-antd/Radio/Radio.tsx
+++ b/src/components/custom-antd/Radio/Radio.tsx
@@ -1,0 +1,7 @@
+import {AntdCustomStyledRadio, ICustomRadioProps} from './Radio.styled';
+
+const Radio: React.FC<ICustomRadioProps> = props => {
+  return <AntdCustomStyledRadio {...props} />;
+};
+
+export default Radio;
--- a/src/components/custom-antd/Radio/RadioGroup/RadioGroup.styled.tsx
+++ b/src/components/custom-antd/Radio/RadioGroup/RadioGroup.styled.tsx
@@ -1,0 +1,5 @@
+import {Radio} from 'antd';
+
+import styled from 'styled-components';
+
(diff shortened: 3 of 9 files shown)


pavloburchak:
Hey, @chooco13!
I have an error happening, when I go to the executors route. It happens to be that sometimes executors don't have image field, so it it required to check this. Could you please fix?
image

chooco13:
I added commit
`handle the case if executor image not exist`

I tested it in my local environment, but please check it one more time.

Thanks for giving me the opportunity to contribute.

pavloburchak:
Perfect! Thank you! From technical side - all good
@fivenp could you please check UX of this feature, please?

fivenp:
Thank you so much for this amazing contribution @chooco13 – very well done :) 

From a user perspective we know that the `container` type will be by far the most used one – so we should def. default to it. 

Since we'd like to keep the executor creation flow as clean and simple as possible I'd propose to move the "Executor type" choice into the settings of an executor (by default it should still be `container` on creation)

image


I do have a higher level question for this though. 
What exactly is the difference between a `job` and `container` executor? What does it imply (also technically)? And why/when would you prefer one over another?

nicufk:
Hey @chooco13, for a dding a new executor of job type you have to follow the instructions from here https://kubeshop.github.io/testkube/test-types/executor-maven#using-different-jdks
it is made for maven but it can be adapted for any other executor

chooco13:
@nicufk 
Are you saying that the user should write the executor only via commands and not from the dashboard?

fivenp:
> @nicufk
> Are you saying that the user should write the executor only via commands and not from the dashboard?

No – I think what @nicufk is referring to is that we will add the Playwright executor as an official executor to our product which will be available to every user out of the box. 

That way there is no need for the user anymore to add a custom executor :)

PR #517, opened 2023-01-26
[#2876] UI: Improve how we show dates and time everywhere

PunkSage:


.eslintrc.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
     'no-useless-escape': 'off',
     'object-curly-newline': 'off',
     'lines-between-class-members': 'off',
-    'unused-imports/no-unused-imports-ts': 'error',
+    'unused-imports/no-unused-imports-ts': 'warn', // should be warned and allowed for compilation
```
pavloburchak, line 32: Why do you think it should be?

src/components/custom-antd/Tooltip/Tooltip.tsx
```diff
@@ -1,3 +1,5 @@
+import React from 'react';
```
pavloburchak, line 1: Is react autoimported for you over here?

.eslintrc.js
PunkSage: When you do the heavy refactoring and removing the pieces of code some functions will become redundant. These imports need to be fixed when creating a PR for sure. However, we should not display an error when the code is perfectly compilable. It breaks focus of the developer, feedback-loop from the IDE and browser downgrading the developer's performance when coding and making changes. If we mark these as warnings these issues could be fixed as a last step.

src/components/custom-antd/Tooltip/Tooltip.tsx
PunkSage: My IDE (Webstorms) complains about relying on globally accessible imports with the UMD. I've seen these imports of React in many places in the codebase. So I guess it is safer to include it. Either way we should be concise to either include it or not in all the files in the codebase.

.eslintrc.js
pavloburchak: Thanks, I like it

src/components/custom-antd/Tooltip/Tooltip.tsx
```diff
@@ -9,7 +11,7 @@ type TooltipProps = {
 const Tooltip: React.FC<AntdTooltipProps & TooltipProps> = props => {
   const {customColor, ...rest} = props;
 
-  return <AntdTooltip color={customColor || Colors.greyBG} {...rest} />;
+  return <AntdTooltip color={customColor || Colors.greyBG} mouseEnterDelay={0.39} mouseLeaveDelay={0.1} {...rest} />;
```
fivenp, line 14: Are we sure this Tooltip isn't used anywhere else? I'm afraid we mess up delays in other places then the dates...

Maybe we should add a new prop which is something like `delay="slow"` which refers to these values for the date tooltips and default to the original setting (which is for sure controlled globally through antd) for everything else?

fivenp (approved):
love it – great job

src/components/custom-antd/Tooltip/Tooltip.tsx
PunkSage: The AntdTooltip in couple of places, but with the changes these values could be set as as defaults.

You can override them you can use mouseEnterDelay, mouseLeaveDelay for the specific Tooltip to behave differently.

If you can provide specific values for these delays for slow, normal, fast then I can follow your advice. 

For know I would leave as it is.

PR #521, opened 2023-01-31
[#3022] Replace Copy button with Download button if network is insecure

PunkSage:
This PR resolves problem: https://github.com/kubeshop/testkube/issues/3022


src/components/molecules/LogOutput/LogOutputActions.tsx
```diff
@@ -16,9 +19,15 @@ type LogOutputActionsProps = {
...
 const LogOutputActions: React.FC<LogOutputActionsProps> = props => {
   const {logOutput, actions} = props;
+  const isSecureContext = useSecureContext();
 
   const logOutputActionsMap: {[key in LogAction]: any} = {
-    copy: <CopyButton content={stripAnsi(logOutput)} key="copy-action" />,
+    copy: isSecureContext ? (
+      <CopyButton content={stripAnsi(logOutput)} key="copy-action" />
+    ) : (
+      // TODO: (Punksage): Find out the name to save the file
```
vsukhin, line 28: should it be dynamic name or based on execution id?

package.json
```diff
@@ -75,6 +76,7 @@
     "axios": "0.27.2",
     "classnames": "2.3.1",
     "date-fns": "^2.28.0",
+    "file-saver": "^2.0.5",
```
vsukhin, line 79: we added only one library, but we have 50k changes in lock file?

src/hooks/useDownloadFile.ts
```diff
@@ -0,0 +1,33 @@
...
+import {saveAs} from 'file-saver';
+
+type DownloadOptions = {
+  filename?: string;
+};
+
+export const useDownloadFile = (textToCopy: string, options: DownloadOptions) => {
+  const [isProcessed, setProcessed] = useState(false);
+
+  const downloadFile = (content: string, filename: string = 'output.sh') => {
```
vsukhin, line 12: same question

src/hooks/useSecureContext.ts
```diff
@@ -0,0 +1,6 @@
+const useSecureContext = () => {
+  // TODO: (Punksage): After testing return it to window.isSecureContext
```
vsukhin, line 2: Is it for testing?

vsukhin (commented):
a few questions

src/components/molecules/LogOutput/LogOutputActions.tsx
PunkSage: @vsukhin I can be either way. How about the format {execution-name}-{hash-of-the-content}.log?

package.json
PunkSage: I'll investigate it. I had a problem with installation of the the packages. I'll reset to the package-lock.json from main and try again with the installation.

src/hooks/useSecureContext.ts
PunkSage: Exactly. This is for the testing. 

If you wish you can trigger the Copy/Download button functionality without changing http-https protocol to server Testkube. I've left it for you so other developers can discover the feature or verify if it is working correctly.

src/hooks/useDownloadFile.ts
PunkSage: I'm open for suggestions. For know I have the format {execution-name}.sh. If you have any preference please let me know.

src/components/atoms/CopyButton/DownloadButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const DownloadButton: React.FC<SaveOptions> = props => {
+  const {content, onClick, filename} = props;
+  const {setProcessed} = useDownloadFile(content, {
+    filename,
+  });
+
+  const onDownloadClick = () => {
+    setProcessed(true);
+    onClick?.();
```
pavloburchak, line 23: I guess it basically replaces ```if (onClick) { onClick(); }```
Never seen this syntax

src/hooks/useCopyToClipboard.ts
```diff
@@ -1,29 +1,37 @@
...
-      }, timeoutInMs);
+    if (isProcessed && isSecureContext) {
+      timeout = copyToClipBoard();
     }
 
     return () => {
       clearTimeout(timeout);
     };
-  }, [isCopied]);
+  }, [isProcessed, window]);
```
pavloburchak, line 34: Just wondering why we need `window` in the dependency array

pavloburchak:
image
I guess you should change the icon, when it is supposed to be downloaded

pavloburchak (commented):
image
You should change copy to download here as well

src/hooks/useCopyToClipboard.ts
PunkSage: To be honest I'm not entirely sure. I think it might be redundant.

src/components/atoms/CopyButton/DownloadButton.tsx
PunkSage: This is the ES6 syntax for optional chaining: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

It will execute the function onClick if it is not undefined. Hence no additional check is required.

src/hooks/useCopyToClipboard.ts
PunkSage: Removed

src/components/atoms/CopyButton/DownloadButton.tsx
pavloburchak: This makes sense, never seen it on methods, but ok

src/components/molecules/CLICommands/CLICommands.tsx
```diff
@@ -36,46 +37,53 @@ const testSuiteScripts: CLIScript[] = [
   {
     label: 'Run test suite',
     command: (name: string) => `kubectl testkube run testsuite ${name}`,
+    filename: 'run-test-suite.sh',
   },
   {
     label: 'Delete test suite',
     command: (name: string) => `kubectl testkube delete testsuite ${name}`,
+    filename: 'delete-test-suite.sh',
```
pavloburchak, line 45: all this props filename should be removed in future, after we create a function that transforms ```This is my test``` into ```this-is-my-test```.
I think we can use it in this case, basically transforming label to filename, adding by default ```.sh``` after.

src/components/molecules/CLICommands/CLICommands.tsx
```diff
@@ -36,46 +37,53 @@ const testSuiteScripts: CLIScript[] = [
...
   {
     label: 'Watch execution',
     command: (name: string) => `kubectl testkube watch execution ${name}`,
-    modify: 'isFinished',
+    filename: 'watch-executions.sh',
   },
   {
     label: 'Download artifacts',
     command: (name: string) => `kubectl testkube download artifacts ${name}`,
-    modify: 'canHaveArtifacts',
```
pavloburchak, line 78: What is a point of removing this modify prop, isn't it used just below?
I can't tell exactly about this block, but as first look, it shouldn't be removed

src/components/pages/Executors/ExecutorDetails/ExecutorSettings/ExecutorSettings.tsx
```diff
@@ -4,10 +4,10 @@ import {SettingsLeftNavigation, StyledSettingsContainer, StyledTabContentContain
 
 import CommandAndArguments from './CommandAndArguments';
 import ContainerImage from './ContainerImage';
-import Definition from './Definition';
+import ExecutorDefinition from './ExecutorDefinition';
 import General from './General';
 
-const tabsConfig: Array<JSX.Element> = [<General />, <ContainerImage />, <CommandAndArguments />, <Definition />];
```
pavloburchak, line 10: What is a point of removing type?

src/components/molecules/CLICommands/CLICommands.tsx
PunkSage: You're right this in accidental change that shouldn't have happened. Reverting back.

src/components/molecules/CLICommands/CLICommands.tsx
PunkSage: We don't need them. Thanks for spotting it. We got the name directly from the url that is generated for the entity.

src/components/pages/Executors/ExecutorDetails/ExecutorSettings/ExecutorSettings.tsx
PunkSage: The type is automatically inferred so there is no reason for typing something that is done automatically by the TypeScript compiler. 

Screenshot 2023-02-03 at 16 47 58

PunkSage:
@pavloburchak Please review the changes. I think the code is mergeable right now.

pavloburchak (approved):

PR #533, opened 2023-02-09
fix: Unable to open old executions

PunkSage:
Resolves: https://github.com/kubeshop/testkube/issues/2962

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,7 +18,7 @@ module.exports = {
     'react-hooks/rules-of-hooks': 'error', // Checks rules of Hooks
     'no-console': 'warn',
     'no-undef': 'off',
-    semi: 'error',
+
     'no-unused-vars': 'off',
     'import/no-extraneous-dependencies': 'off',
     'prefer-destructuring': 'off',
@@ -29,6 +29,7 @@ module.exports = {
     'no-useless-escape': 'off',
     'object-curly-newline': 'off',
     'lines-between-class-members': 'off',
+    semi: 'warn', // should be warned and allowed for compilation
     'no-restricted-globals': 'warn', // should be warned and allowed for compilation
     'unused-imports/no-unused-imports-ts': 'warn', // should be warned and allowed for compilation
     'react/jsx-curly-brace-presence': 'warn', // should be warned and allowed for compilation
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
     'react-hooks/rules-of-hooks': 'error', // Checks rules of Hooks
     'no-console': 'warn',
     'no-undef': 'off',
-    semi: 'error',
```
PunkSage, line 21: All of these problems that allow for proper compilation should not prevent a developer from seeing the results of the changes

src/components/organisms/EntityDetails/EntityDetailsContainer/EntityDetailsContainer.tsx
```diff
@@ -177,28 +177,15 @@ const EntityDetailsContainer: React.FC<EntityDetailsBlueprint> = props => {
   };
 
   useEffect(() => {
-    if (execId && executionsList && executionsList?.results) {
```
Toxer10, line 180: Does this feature work without this code?

.eslintrc.js
Toxer10: I would not want to have ```warn``` on this rule, it's too crucial to be omitted ocassionally

Toxer10 (changes requested):

.eslintrc.js
PunkSage: All of these issues could be marked as warnings and be fixed when you compile and push-out the code to the repository. Throwing this as an error breaks the developers feedback loop and disallow you do concentrate on the root of the problem you’re trying to solve. After a few changes of the code you are forced to fight with the linter.

src/components/organisms/EntityDetails/EntityDetailsContainer/EntityDetailsContainer.tsx
PunkSage: I cannot find any feature that would be broken after removal of this code. If you can verify this I would greatly appreciate.

src/components/organisms/EntityDetails/EntityDetailsContainer/EntityDetailsContainer.tsx
```diff
@@ -177,28 +177,24 @@ const EntityDetailsContainer: React.FC<EntityDetailsBlueprint> = props => {
...
+      const indexOfDisplayedExecution = executionDetails ? executionsList.results?.indexOf(executionDetails) + 1 : null;
+      if (!indexOfDisplayedExecution) setDaysFilterValue(0);
+    }
+  }, [params, executionsList]);
 
-        const targetEntityArrayIndex = executionsList.results?.indexOf(targetEntity[0]) + 1;
+  useEffect(() => {
+    const executionDetails = executionsList?.results?.find((execution: any) => execution.id === execId);
+    const indexOfDisplayedExecution = executionDetails ? executionsList.results?.indexOf(executionDetails) + 1 : null;
+    if (indexOfDisplayedExecution) setCurrentPage(Math.ceil(indexOfDisplayedExecution / 10));
```
pavloburchak, line 190: Here the same as above

pavloburchak (approved):

PR #561, opened 2023-03-23
feat: add loading indicator to button on click

haneabogdan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/EntityDetails/EntityDetailsContent/EntityDetailsContent.tsx
+++ b/src/components/organisms/EntityDetails/EntityDetailsContent/EntityDetailsContent.tsx
@@ -15,6 +15,7 @@ import {Button, Text} from '@custom-antd';
 
 import {CLICommands, LabelsList, MetricsBarChart, notificationCall} from '@molecules';
 
+import useLoadingIndicator from '@hooks/useLoadingIndicator';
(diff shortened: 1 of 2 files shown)


src/components/organisms/EntityDetails/EntityDetailsContent/EntityDetailsContent.tsx
```diff
@@ -74,6 +76,7 @@ const EntityDetailsContent: React.FC = () => {
   const testIcon = entityDetails?.testIcon;
 
   const onRunButtonClick = async () => {
```
rangoo94, line 78: Shouldn't it instead set the loading status based on `runEntity` result in `onRunButtonClick`?

haneabogdan: there's the possibility to access the loading state of the request using the "isLoading" flag provided directly by the RTK mutation. tried this first, but that's not enough, as the request is resolved but the agent is still running the test. If there are too many tests running in parallel, the agent will stop responding and return 503s. the agreement was, to have a "fixed" 2s window for this. The issue is impacting cloud especially.

rangoo94 (approved):

PR #562, opened 2023-03-24
Remove fixed filters and add infinite scroll indicator

rangoo94:
## Changes

- Make the filters scroll along with the entities list
- Move "Infinite Scroll" logic outside of the `GridItem` to the top level
   - Create `ScrollTarget` component that helps to perform that
- Add a loading icon as an indicator for loading the next entities

## Fixes

- https://github.com/kubeshop/testkube/issues/3025

## How to test it

- Scroll on "Tests" or "Test Suites" list

## screenshots

Zrzut ekranu 2023-03-24 o 13 01 06

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [x] added a test


fivenp:
@rangoo94 – can we please also remove the stickiness from the title (there's no real benefit in having that sticky) and just make the whole container scrollable? :)

rangoo94:
> @rangoo94 – can we please also remove the stickiness from the title (there's no real benefit in having that sticky) and just make the whole container scrollable? :)

Thanks @fivenp, I will update the ticket and apply this change.

Additionally, @pavloburchak has requested the loading indicator to be bigger, and it makes sense, so will do it as well 👍

PR #575, opened 2023-04-04
Pavlo/git commit

pavloburchak:
This PR...

## Changes
- refactored all test creation and test edit logic
- https://github.com/kubeshop/testkube/issues/3583
- https://github.com/kubeshop/testkube/issues/3157

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


package.json
```diff
@@ -57,6 +57,7 @@
     "json-server": "0.17.0",
     "less": "4.1.2",
     "log-update": "5.0.1",
+    "luxon": "^3.2.1",
```
rangoo94, line 60: Do you actually use `luxon` somewhere, or `package.json`/`package-lock.json` changes are not needed?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "testkube-dashboard",
-  "version": "1.9.0",
+  "version": "1.10.4",
```
rangoo94, line 3: Should we update the version? if so, maybe it's better in a separate PR?

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
```diff
@@ -0,0 +1,183 @@
...
+        <FormItem
+          name="name"
+          label="Name"
+          rules={[required, k8sResourceNamePattern, k8sResourceNameMaxLength]}
+          required
+        >
+          <Input placeholder="e.g.: my-test" />
+        </FormItem>
+        <FormItem name="testType" label="Type" rules={[required]} required>
+          <Select placeholder="Select from available executors..." showSearch options={remappedExecutors} allowClear />
```
rangoo94, line 95: Do we want to `allowClear`?

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
```diff
@@ -0,0 +1,183 @@
...
+        </FormItem>
+        <FormItem
+          noStyle
+          shouldUpdate={(prevValues, currentValues) => prevValues.testSource !== currentValues.testSource}
+        >
+          {({getFieldValue}) => {
+            const selectedExecutor = executors.find((executor: Executor) =>
```
rangoo94, line 102: Maybe it's worth putting it on top, using the `form` variable and `useMemo`?

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
```diff
@@ -0,0 +1,183 @@
...
+            if (!selectedExecutor) {
+              return null;
+            }
+
+            const options = [
+              ...remappedCustomTestSources,
+              ...testSourceBaseOptions.filter(
+                option =>
+                  selectedExecutor.executor?.contentTypes?.includes(String(option.value)) ||
+                  !selectedExecutor.executor?.contentTypes?.length
```
rangoo94, line 115: Is that correct that when the executor doesn't specify types it should allow any? Makes sense for me, but I'd like to confirm.

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
```diff
@@ -0,0 +1,183 @@
...
+              ),
+            ];
+
+            return (
+              <FormItem rules={[required]} label="Source" name="testSource" required>
+                <Select placeholder="Select a source..." options={options} showSearch allowClear />
```
rangoo94, line 121: Should we `allowClear`?

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
```diff
@@ -0,0 +1,183 @@
...
+          noStyle
+          shouldUpdate={(prevValues, currentValues) => prevValues.testSource !== currentValues.testSource}
+        >
+          {({getFieldValue}) => {
+            const testSourceValue = getSourceFieldValue(getFieldValue);
+
+            const selectedExecutor = executors.find((executor: any) =>
+              executor.executor?.types?.includes(getFieldValue('testType'))
+            );
+            const executorType = selectedExecutor?.executor.meta?.iconURI;
```
rangoo94, line 136: If the `selectedExecutor` would be on top, this could be extracted as well

src/components/organisms/TestConfigurationForm/molecules/BranchCommitSwitcher.tsx
rangoo94, line 1: Instead of doing it this way, couldn't we rename field to `revision`? It would match both branch and commit, while Git commands are exactly the same to checkout this code

src/utils/executors.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  executors.forEach(executorItem => {
+    const {
+      executor: {types, executorType},
+    } = executorItem;
+
+    if (types) {
+      types.forEach(type => {
+        array.push({label: type, value: type});
+      });
```
rangoo94, line 19: What do you think about i.e.:

```suggestion
      array.push(...types.map(type => ({label: type, value: type})));
```

And similarly below?

src/utils/sources.ts
```diff
@@ -0,0 +1,159 @@
...
+
+import {notificationCall} from '@molecules';
+
+const customGitSourceString = '$custom-git-';
+
+export const remapTestSources = (testSources: SourceWithRepository[]) => {
+  if (!testSources || !testSources.length) {
+    return [];
+  }
+
```
rangoo94, line 14: Based on the function definition, `testSources` is always an array, so this part could be omitted.

```suggestion
```

src/utils/sources.ts
```diff
@@ -0,0 +1,159 @@
...
+      secrets.tokenSecret = {};
+    } else if (!token.includes('*')) {
+      secrets.token = token;
+    }
+  }
+
+  if (username !== undefined) {
+    if (username === '') {
+      secrets.usernameSecret = {};
+    } else if (!username.includes('*')) {
```
rangoo94, line 65: What if someone actually puts `*` in a username or token - shouldn't it instead show a validation error?

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
pavloburchak: Yes, it is

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
pavloburchak: No particular reason to have to. Do you think it should be removed?

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
pavloburchak: Agree, will move it

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
rangoo94: I can't find a case where any User would need to clear it, and we will have to ensure that this works well. I would suggest disallowing clearing - it's always easier to maintain a feature that doesn't exist 😄

src/components/organisms/TestConfigurationForm/molecules/Revision.tsx
```diff
@@ -0,0 +1,34 @@
...
+      {({getFieldValue}) => {
+        const commit = getFieldValue('commit');
+
+        if (isFirst) {
+          setSwitcherValue(commit ? 'commit' : 'branch');
+        }
+
+        return (
+          <StyledFormSpace size={24} direction="vertical">
+            <RevisionSwitcher value={switcherValue} onChange={setSwitcherValue} />
```
rangoo94, line 25: By changing to `Revision`, I didn't mean component name - I meant, that we could get rid of `Commit`, `Branch`, and `RevisionSwitcher` and just have `Revision` - simple text input with `Git revision` label.

It doesn't matter whether the user will provide a branch name or commit hash. Git commands will operate on any:

```
git checkout main
git checkout HEAD
git checkout 23e55a1449cf0bcc4b8e6cd7fbcdfbd97d2e662c
```

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
```diff
@@ -0,0 +1,180 @@
...
+              ),
+            ];
+
+            return (
+              <FormItem rules={[required]} label="Source" name="testSource" required>
+                <Select placeholder="Select a source..." options={options} showSearch />
+              </FormItem>
+            );
+          }}
+        </FormItem>
```
rangoo94, line 125: I see that there is no `Branch/Commit`, which is in the edition. Should it be there?

src/components/organisms/TestConfigurationForm/molecules/Revision.tsx
rangoo94: Just tested, and for `commit` property at the moment you may pass either commit hash or branch name and it works, so it behave as `revision` would :)

src/components/organisms/TestConfigurationForm/molecules/RevisionSwitcher.tsx
rangoo94, line 1: Noticed here, but apply to many files - should the `organism` have these molecules? Shouldn't molecules be flattened to `src/components/molecules`?

rangoo94 (commented):
Looks like it will also solve https://github.com/kubeshop/testkube/issues/3410 afterwards

src/components/organisms/TestConfigurationForm/molecules/Revision.tsx
rangoo94: @vsukhin, what do you think about `revision` approach?

pavloburchak: Oh, I see. I am not sure about it. Probably need to discuss with the team

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
pavloburchak: Nope, for test creation it is only a branch. Possibility to select input a commit only available in edition

src/components/organisms/TestConfigurationForm/molecules/Path.tsx
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const Path: React.FC<PathProps> = props => {
+  const {testType} = props;
+
+  return (
+    <FormItem
+      name="path"
+      label="Path"
+      rules={[required]}
```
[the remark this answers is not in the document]
pavloburchak: Yep, thanks

src/components/organisms/EntityList/EntityCreationModal/TestCreationForm/TestCreationForm.tsx
```diff
@@ -0,0 +1,180 @@
...
+              ...remappedCustomTestSources,
+              ...testSourceBaseOptions.filter(
+                option =>
+                  selectedExecutor.executor?.contentTypes?.includes(String(option.value)) ||
+                  !selectedExecutor.executor?.contentTypes?.length
+              ),
+            ];
+
+            return (
+              <FormItem rules={[required]} label="Source" name="testSource" required>
```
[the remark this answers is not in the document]
pavloburchak: Actually yes, it was present before. Quiestion is more about priority, when test is already have branch and path, what is more important to prefill, custom sources fields or test?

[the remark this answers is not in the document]
rangoo94: I think that the tests data should have a priority

src/utils/sources.ts
```diff
@@ -0,0 +1,157 @@
...
+};
+
+export const getCustomSourceField = (testSource: string, prevTestSource?: string) => {
+  const isCustomTestSource = testSource.includes(customGitSourceString);
+
+  return isCustomTestSource
+    ? {source: testSource.replace(customGitSourceString, '')}
+    : prevTestSource
+    ? {source: ''}
+    : {};
```
rangoo94, line 106: Probably in this case it could be always `source: ''`

rangoo94 (approved):

PR #577, opened 2023-04-05
external link component

pavloburchak:
This PR...

## Changes
- added External link component
- verify it is used in all places and behaviour it the same

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/EndpointModal/EndpointModal.tsx
+++ b/src/components/molecules/EndpointModal/EndpointModal.tsx
@@ -2,13 +2,15 @@ import React, {useContext, useEffect, useState} from 'react';
 
 import {Space} from 'antd';
 
+import {ExternalLink} from '@atoms';
+
 import {Button, Input, Modal, Text} from '@custom-antd';
+
+import {useApiEndpoint, useUpdateApiEndpoint} from '@services/apiEndpoint';
 
 import Colors from '@styles/Colors';
 
 import {MainContext} from '@contexts';
-
-import {useApiEndpoint, useUpdateApiEndpoint} from '@services/apiEndpoint';
 
 import notificationCall from '../Notification/Notification';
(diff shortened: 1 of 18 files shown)


src/components/molecules/EndpointModal/EndpointModal.tsx
```diff
@@ -59,13 +61,9 @@ const EndpointModal: React.FC<EndpointModalProps> = props => {
           <Text>
             We could not detect the right Testkube API endpoint for you. Please enter the API endpoint for your
             installation (e.g. from the output of the Testkube installer)&nbsp;
-            <a
-              href="https://kubeshop.github.io/testkube/UI/#ui-results-endpoint"
-              target="_blank"
-              rel="noopener noreferrer"
-            >
+            <ExternalLink href="https://kubeshop.github.io/testkube/UI/#ui-results-endpoint">
```
rangoo94, line 64: Missing `rel="noopener noreferrer"`

rangoo94 (approved):
I don't think that such component is needed (for me `<a target='_blank'>` is fine and simpler to customize), but if you prefer to use it instead of `a[target=_blank]` it's fine for me :)

src/components/molecules/EndpointModal/EndpointModal.tsx
pavloburchak: it is not needed here

pavloburchak:
> I don't think that such component is needed (for me `<a target='_blank'>` is fine and simpler to customize), but if you prefer to use it instead of `a[target=_blank]` it's fine for me :)

I would like to extend it in future. And for me it is just more intuitive to customise it

PR #585, opened 2023-04-17
fix: update styles to conform to Cloud more

rangoo94:
## Changes

- It changes few small styles, so the OSS styles and code will more match the Cloud one

## Fixes

- part of https://github.com/kubeshop/testkube/issues/3695

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


rangoo94:
@fivenp, could you check if that's fine, or maybe should it be moved the other way around - from OSS to Cloud?

PR #591, opened 2023-04-18
chore: use PropsWithChildren instead of manually typing `children` property

rangoo94:
## Changes

- same as in Cloud

## Fixes

- Part of https://github.com/kubeshop/testkube/issues/3695

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


pavloburchak:
I think that children property is also should be removed in `molecules/Definition` component

rangoo94:
> I think that children property is also should be removed in `molecules/Definition` component

Thanks @pavloburchak, added 👍

PR #595, opened 2023-04-18
feat: Repository Validation

pavloburchak:
This PR...

## Changes
https://github.com/kubeshop/testkube/issues/3319
- repo validation on test creation and test edit

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/antd-theme/my-antd-theme.less
+++ b/src/antd-theme/my-antd-theme.less
@@ -68,3 +68,5 @@
 @segmented-selected-bg: #818CF8;
 @segmented-label-color: #FFFFFF;
 @segmented-label-hover-color: #FFFFFF;
+// form
+@label-required-color: #F43F5E;
--- a/src/components/molecules/GitFormItems/tooltipUtils.tsx
+++ b/src/components/molecules/GitFormItems/tooltipUtils.tsx
@@ -11,7 +11,7 @@ export const tooltipIcons: {[key in tooltipStatus]: JSX.Element | undefined} = {
   none: undefined,
 };
 
-export const getTooltip = (status: tooltipStatus, message?: string) => {
+export const getValidationTooltip = (status: tooltipStatus, message?: string) => {
   if (!tooltipIcons[status]) {
     return undefined;
   }
--- a/src/components/molecules/GitFormItems/tooltipUtils.tsx
+++ b/src/components/molecules/GitFormItems/tooltipUtils.tsx
@@ -1,0 +1,24 @@
+import {CheckCircleFilled, LoadingOutlined, WarningFilled} from '@ant-design/icons';
+
+import Colors from '@styles/Colors';
+
+export type tooltipStatus = 'loading' | 'error' | 'success' | 'none';
+
+export const tooltipIcons: {[key in tooltipStatus]: JSX.Element | undefined} = {
+  loading: <LoadingOutlined />,
+  success: <CheckCircleFilled style={{color: Colors.lime400, cursor: 'auto', paddingBottom: '2px'}} />,
+  error: <WarningFilled style={{color: Colors.amber400}} />,
+  none: undefined,
+};
+
+export const getTooltip = (status: tooltipStatus, message?: string) => {
+  if (!tooltipIcons[status]) {
+    return undefined;
+  }
+
+  if (status === 'error') {
+    return {title: message, icon: tooltipIcons[status]};
+  }
+
+  return {icon: tooltipIcons[status]};
+};
--- a/src/components/molecules/GitFormItems/tooltipUtils.tsx
+++ b/src/components/molecules/GitFormItems/tooltipUtils.tsx
@@ -2,16 +2,21 @@ import {CheckCircleFilled, LoadingOutlined, WarningFilled} from '@ant-design/ico
 
 import Colors from '@styles/Colors';
 
-export type tooltipStatus = 'loading' | 'error' | 'success' | 'none';
+export enum TooltipStatus {
+  loading = 'loading',
+  error = 'error',
+  success = 'success',
+  none = 'none',
+}
 
-export const tooltipIcons: {[key in tooltipStatus]: JSX.Element | undefined} = {
+export const tooltipIcons: Record<TooltipStatus, JSX.Element | undefined> = {
   loading: <LoadingOutlined style={{cursor: 'auto'}} />,
   success: <CheckCircleFilled style={{color: Colors.lime400, cursor: 'auto', paddingBottom: '2px'}} />,
   error: <WarningFilled style={{color: Colors.amber400, paddingBottom: '2px'}} />,
   none: undefined,
 };
 
-export const getValidationTooltip = (status: tooltipStatus, message?: string) => {
+export const getValidationTooltip = (status: TooltipStatus, message?: string) => {
   if (!tooltipIcons[status]) {
     return undefined;
   }
--- a/src/components/molecules/GitFormItems/tooltipUtils.tsx
+++ b/src/components/molecules/GitFormItems/tooltipUtils.tsx
@@ -5,9 +5,9 @@ import Colors from '@styles/Colors';
 export type tooltipStatus = 'loading' | 'error' | 'success' | 'none';
 
 export const tooltipIcons: {[key in tooltipStatus]: JSX.Element | undefined} = {
-  loading: <LoadingOutlined />,
+  loading: <LoadingOutlined style={{cursor: 'auto'}} />,
   success: <CheckCircleFilled style={{color: Colors.lime400, cursor: 'auto', paddingBottom: '2px'}} />,
-  error: <WarningFilled style={{color: Colors.amber400}} />,
+  error: <WarningFilled style={{color: Colors.amber400, paddingBottom: '2px'}} />,
   none: undefined,
 };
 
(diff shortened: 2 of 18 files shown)


src/antd-theme/my-antd-theme.less
```diff
@@ -67,4 +67,6 @@
 @segmented-hover-bg: #1E293B;
 @segmented-selected-bg: #818CF8;
 @segmented-label-color: #FFFFFF;
-@segmented-label-hover-color: #FFFFFF;
\ No newline at end of file
+@segmented-label-hover-color: #FFFFFF;
+// form
+@label-required-color: #F43F5E;
```
rangoo94, line 72: Missing empty line on the end of file

src/components/molecules/GitFormItems/tooltipUtils.tsx
```diff
@@ -0,0 +1,24 @@
+import {CheckCircleFilled, LoadingOutlined, WarningFilled} from '@ant-design/icons';
+
+import Colors from '@styles/Colors';
+
+export type tooltipStatus = 'loading' | 'error' | 'success' | 'none';
```
rangoo94, line 5: Better to use `enum`:
```suggestion
enum TooltipStatus {
  loading = 'loading',
  error = 'error',
  success = 'success',
  none = 'none',
}
export type tooltipStatus = 'loading' | 'error' | 'success' | 'none';
```

src/hooks/useValidateRepository.ts
```diff
@@ -0,0 +1,188 @@
...
+const endSearchString = ', context';
+
+const getErrorMessage = (rawErrorString: string, searchString: string) =>
+  rawErrorString.slice(
+    rawErrorString.search(searchString) + searchString.length,
+    rawErrorString.search(endSearchString)
+  );
+
+const useValidateRepository = (
+  getFieldValue: (name: NamePath) => string,
```
rangoo94, line 23: Why `getFieldValue` instead of the values object?

src/hooks/useValidateRepository.ts
```diff
@@ -0,0 +1,188 @@
...
+
+const fieldsNames: string[] = ['uri', 'token', 'username', 'branch', 'path'];
+
+const authSearchString = 'username or token:';
+const uriSearchString = 'uri:';
+const branchSearchString = 'branch:';
+const pathSearchString = 'path:';
+const endSearchString = ', context';
+
+const getErrorMessage = (rawErrorString: string, searchString: string) =>
```
rangoo94, line 16: Shouldn't back-end return map of errors for each field separately instead?

src/hooks/useValidateRepository.ts
```diff
@@ -0,0 +1,188 @@
...
+  rawErrorString.slice(
+    rawErrorString.search(searchString) + searchString.length,
+    rawErrorString.search(endSearchString)
```
rangoo94, line 19: `String.prototype.search` is using `RegExp`. There you want to search for the actual string, so `indexOf` is more appropriate:
```suggestion
    rawErrorString.indexOf(searchString) + searchString.length,
    rawErrorString.indexOf(endSearchString)
```

Overall, parsing the message string feels hacky a lot, so most likely it will have some bugs (i.e. because of`endSearchString` found before `searchString`)

src/hooks/useValidateRepository.ts
```diff
@@ -0,0 +1,188 @@
...
+          username: getFieldValue('username') ? 'success' : '',
+          branch: getFieldValue('branch') ? 'success' : '',
+          path: getFieldValue('path') ? 'success' : '',
+        });
+      }
+    },
+    [setValidationState, getFieldValue]
+  );
+
+  const getValidationPayload = useCallback(() => {
```
rangoo94, line 97: It doesn't have to be wrapped in `useCallback`, as it's not passed anywhere as dependency (no strict equal check)

src/hooks/useValidateRepository.ts
```diff
@@ -0,0 +1,188 @@
...
+    rawErrorString.search(searchString) + searchString.length,
+    rawErrorString.search(endSearchString)
+  );
+
+const useValidateRepository = (
+  getFieldValue: (name: NamePath) => string,
+  setValidationState: React.Dispatch<any>,
+  validateRepository: any
+) => {
+  const handleResponse = useCallback(
```
rangoo94, line 27: It doesn't have to be wrapped in `useCallback`, as it's not passed anywhere as dependency (no strict equal check)

src/hooks/useValidateRepository.ts
```diff
@@ -0,0 +1,188 @@
...
+      }
+
+      setValidationState({
+        uri: 'loading',
+        token: getFieldValue('token') ? 'loading' : '',
+        username: getFieldValue('username') ? 'loading' : '',
+      });
+
+      validateRepository(getValidationPayload()).then((res: any) => {
+        handleResponse(res);
```
rangoo94, line 132: After the response is there, it should check if the values are the same. Otherwise, it will fall into a race condition:

- 0ms - User updates the data **(invalid)**
- 300ms - Started 1st validation request
- 400ms - User updates the data **(valid)**
- 700ms - Started 2nd validation request
- 900ms - 2nd validation request finished and the validation results were updated
- 1000ms - 1st validation request finished **and the validation results were incorrectly updated to invalid**

src/hooks/useValidateRepository.ts
```diff
@@ -0,0 +1,188 @@
...
+      });
+    },
+    300,
+    [getFieldValue('uri'), getFieldValue('token'), getFieldValue('username')]
+  );
+
+  useDebounce(
```
rangoo94, line 139: Shouldn't all these 3 debounces be combined together? These are done for exactly same thing, and each of them is making the request separately

pavloburchak: I think they shouldn't when user inputs branch, for example, we shouldn't show loading status on all fields, it should be just a branch. So I combined auth stuff in one debounce, and each other field should have its own one

src/hooks/useValidateRepository.ts
pavloburchak: Sorry, didn't quite get it. Which values and how it is supposed to be checked?

src/hooks/useValidateRepository.ts
pavloburchak: Accepted indexOf change, will talk about this parsing with Vlad

src/hooks/useValidateRepository.ts
pavloburchak: I don't know, I guess this also works

src/hooks/useValidateRepository.ts
pavloburchak: No particular reason, do you think using values object is better?

rangoo94: On the one hand, it may be simpler and more type-safe, on the other - it may be harder to compare them. Not sure what should we prefer, so will mark as resolved

src/hooks/useValidateRepository.ts
rangoo94: The form values. To show different example:

* You set the branch to `branch-1`, API Call #1 is started
* You set the branch to `branch-2`, API Call #2 is started
* API Call #2 is finished - it will show (correctly) results for the latest `branch-2`
* API Call #1 is finished - it will replace the correct results, with results for the old `branch-1`

To avoid that, you need to either cancel the previous API call after changes or ignore it afterward (by i.e. checking if the values are still the same)

src/hooks/useValidateRepository.ts
rangoo94: There are two problems with that:

* API response gives results for all fields at once
* The fields are strictly connected - if you will change the repository URI, it means that the branch may become invalid

Additionally, you don't have to show a loader for all of them, but still you might aggregate them in a single call

src/hooks/useValidateRepository.ts
pavloburchak: Ok, I got it. Thanks

src/hooks/useValidateRepository.ts
pavloburchak: Honestly don't see that much of a problem. Yes, fields are connected, but still branch validation will be correctly displayed if URI is invalid. I will try to see if I can resolve loaders. Not sure right now how to identify which exact field was changed and display loader for it.

rangoo94: If it will be hard - ignore it; this part, as long as the race condition will be solved, should mostly work okay anyway (except unnecessary calls and sometimes some flaky interface)

src/hooks/useValidateRepository.ts
pavloburchak: Resolving comment because solution is ok for current BE implementation

src/hooks/useValidateRepository.ts
```diff
@@ -0,0 +1,189 @@
...
+        username: current.username ? 'success' : '',
+        branch: current.branch ? 'success' : '',
+        path: current.path ? 'success' : '',
+      });
+    }
+  };
+
+  useDebounce(
+    () => {
+      // Can't validate repository without URI or branch name
```
rangoo94, line 149:
```suggestion
      // Can't validate repository without URI
```

rangoo94 (approved):

PR #619, opened 2023-04-24
feat: simplify composing providers, add ModalContext, use it same way as in Cloud

rangoo94:
## Changes

- simplify composing providers, so the Git diffs will be much smaller
- extract global modal mechanism to `ModalContext`
- use it for same feature as in Cloud version (in future it could be used in all such situations)

## Fixes

- part of https://github.com/kubeshop/testkube/issues/3695

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -23,6 +23,7 @@ import {Sider} from '@organisms';
 import {EndpointProcessing, ErrorBoundary, NotFound} from '@pages';
 
 import {PollingIntervals} from '@utils/numbers';
+import {composeProviders} from '@utils/composeProviders';
 
 import {ReactComponent as LoadingIcon} from '@assets/loading.svg';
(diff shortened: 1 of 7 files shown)


pavloburchak:
I think we can delete `models/rtk.js` file

pavloburchak:
Should we use that Modal context everywhere, or this is part of another scope?

rangoo94:
> Should we use that Modal context everywhere, or this is part of another scope?

For now, I only migrated the usage from Cloud, to avoid the complexity of a new code. I think though that it's worth using it in the future - it should rather work with all modals (except nested).

pavloburchak (approved):

PR #623, opened 2023-04-25
Upgrade Node.js to v20

rangoo94:
## Changes

- Upgrade Node.js to v20 and force it
- Avoid `npm-force-resolutions` script as it's not needed (used [**overrides**](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides) instead)
- Use `--openssl-legacy-provider`, as it's needed for running the Webpack 4.x we're using ([**more here**](https://github.com/webpack/webpack/issues/14532))

## Fixes

- https://github.com/kubeshop/testkube/issues/3726

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


rangoo94:
Unfortunately, I don't have access to Vercel to check what happened here - maybe it requires some configuration. @haneabogdan, do you have access there to check?

Toxer10 (approved):

haneabogdan (approved):

PR #625, opened 2023-04-26
refactor: no more warning code in repo

Toxer10:
image


package.json
```diff
@@ -111,9 +111,9 @@
     "test:coverage": "craco test --coverage --watchAll=false",
     "test:update": "craco test --updateSnapshot",
     "eject": "react-scripts eject",
-    "lint": "npm run lint:config && npm run lint:src ",
+    "lint": "npm run lint:config && npm run lint:src",
     "lint:config": "eslint .",
```
rangoo94, line 115: Probably could be the same for `config`

rangoo94 (commented):
It looks good for me. @pavloburchak, @haneabogdan - any objections? :)

package.json
Toxer10: OK, let's add it also there. Did not add it at first because we don't usually use `lint:config` directly. But it's not a big deal

rangoo94: We use `lint` which calls it, so it would not have `--max-warnings=0` otherwise

haneabogdan (approved):

PR #653, opened 2023-05-11
fix: use Git tokens and Git usernames correctly for the Test Sources

rangoo94:
## Changes

- Fix updating the token/username (use `SecretFormItem`, same as for tests, and adjust the payloads)
- Fix creating a test source with token/username (adjust the payloads)
- Fix displaying Test Source's Git URI (it was not refreshing)

## Fixes

- https://github.com/kubeshop/testkube/issues/3810

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/pages/Sources/SourceDetails/SourceSettings/General/Authentication.tsx
+++ b/src/components/pages/Sources/SourceDetails/SourceSettings/General/Authentication.tsx
@@ -1,16 +1,13 @@
-import {useContext} from 'react';
+import {useContext, useEffect, useState} from 'react';
 
-import {Input as AntdInput, Form} from 'antd';
-
-import {EyeInvisibleOutlined, EyeOutlined} from '@ant-design/icons';
+import {Form} from 'antd';
 
 import {useAppSelector} from '@redux/hooks';
 import {selectCurrentSource, setCurrentSource} from '@redux/reducers/sourcesSlice';
 
-import {Input} from '@custom-antd';
+import {ConfigurationCard, notificationCall, SecretFormItem} from '@molecules';
 
-import {ConfigurationCard, notificationCall} from '@molecules';
-
+import {dummySecret} from '@utils/sources';
 import {displayDefaultNotificationFlow} from '@utils/notification';
 
 import {useUpdateSourceMutation} from '@services/sources';
@@ -18,6 +15,8 @@ import {useUpdateSourceMutation} from '@services/sources';
 import {Permissions, usePermission} from '@permissions/base';
 
 import {MainContext} from '@contexts';
+
+import {StyledSpace} from '../../SourceDetails.styled';
 
 type AuthenticationFormValues = {
   token: string;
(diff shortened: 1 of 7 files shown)


src/services/sources.ts
```diff
@@ -28,7 +27,7 @@ export const sourcesApi = createApi({
         body,
       }),
     }),
-    createSource: builder.mutation<MetadataResponse<{name: string; namespace?: string}>, SourceWithRepository>({
+    createSource: builder.mutation<CreateSourceResult, SourceWithRepository>({
```
Toxer10, line 30: I would prefer to use it like this:

1. Change `MetadataResponse` to:
```
export type MetadataResponse<Meta = {}, Spec = {}> = {
  metadata: Meta;
  spec?: Spec;
};
```

2. Use it as:
```
MetadataResponse<
  {
    name: string;
    namespace: string;
    uid: string;
    resourceVersion: string;
    generation: 1;
    creationTimestamp: string;
  },
  {repository: Repository}
>
```

haneabogdan (approved):

src/services/sources.ts
```diff
@@ -53,7 +52,7 @@ export const sourcesApi = createApi({
         method: 'DELETE',
       }),
     }),
-    updateSource: builder.mutation<void, SourceWithRepository>({
+    updateSource: builder.mutation<any, SourceWithRepository>({
```
Toxer10, line 55: [It](https://docs.testkube.io/openapi/#tag/test-sources/operation/updateTestSource) returns an updated `SourceWithRepository`, no? There was `void` because we did not use this updated `SourceWithRepository` after the request.

src/components/pages/Sources/SourceDetails/SourceSettings/General/Authentication.tsx
```diff
@@ -81,16 +96,22 @@ const Authentication: React.FC = () => {
...
+            setIsClearedValue={setIsClearedUsername}
+          />
+          <SecretFormItem
+            name="token"
+            label="Git token"
+            isClearedValue={isClearedToken}
+            setIsClearedValue={setIsClearedToken}
           />
-        </Form.Item>
+        </StyledSpace>
```
Toxer10, line 114: Use here `FullWidthSpace`?

Toxer10 (changes requested):

src/services/sources.ts
rangoo94: `MetadataResponse` should have only metadata, not the `spec`. It could be optionally used though like:
```
MetadataResponse<{
    name: string;
    namespace: string;
    uid: string;
    resourceVersion: string;
    generation: 1;
    creationTimestamp: string;
  }> & {spec: { repository: Repository }}
```

But I don't think that it makes sense, given that it's just wrapping in `meta`

src/services/sources.ts
rangoo94: It doesn't - it returns partial of `spec` and some changes. Didn't want to try to type it, because I'm not sure if it's expected for that.

Toxer10: OK, so we can't relay on docs all the time 👌

src/services/sources.ts
rangoo94: Actually, I see that the `spec` is already used anyway in `MetadataResponse`. Will use it then, although I think that `MetadataResponse` should be rethinked later, so it will be helpful 👍

Toxer10: Let's leave it for now then. I will create smth like `MetaSpecResponse` with 2 arguments for generic `Meta` and `Spec`. But I don't think it's a big problem to have `meta` and `spec` fields typed in `MetadataResponse` as those 2 fields are always come together.

Toxer10 (approved):

src/services/sources.ts
rangoo94: @Toxer10, basically, this `MetadataResponse` is `Definition` :) Because it's CRD's object definition. It doesn't have to have `metadata`, it might be custom, and same for spec.

PR #668, opened 2023-05-15
refactor: changed theme color and styling

Toxer10:
Now disabled button (1) looks better and consistant with `Form.Item`s, `CopyCommand` has more space for command (2) and custom `FormItem` now has more gereral styles.

For 1: image
For 2: 
Prev: 
image

Now:
image

### DIFF
--- a/src/components/atoms/Pre/Pre.styled.tsx
+++ b/src/components/atoms/Pre/Pre.styled.tsx
@@ -1,7 +1,7 @@
 import styled from 'styled-components';
 
 export const StyledPre = styled.pre`
-  padding: 12px 40px 12px 12px;
+  padding: 12px 0 12px 12px;
   margin-bottom: 0;
   border-radius: 4px;
 
--- a/src/components/custom-antd/Form/Form.styled.tsx
+++ b/src/components/custom-antd/Form/Form.styled.tsx
@@ -1,8 +1,18 @@
+import {Form} from 'antd';
+
 import styled from 'styled-components';
 
 import Fonts from '@styles/Fonts';
 
+const {Item: AntdFormItem} = Form;
+
 export const StyledFormItemLabel = styled.label`
   font-family: ${Fonts.nunito};
   font-weight: 300;
+`;
+
+export const StyledFormItem = styled(AntdFormItem)`
+  flex: 1;
+
+  margin-bottom: 0;
 `;
--- a/src/antd-theme/my-antd-theme.less
+++ b/src/antd-theme/my-antd-theme.less
@@ -15,8 +15,8 @@
 @btn-default-color: #cbd5e1;
 @btn-default-border: transparent;
 @btn-disable-color: #64748b;
-@btn-disable-bg: #334155;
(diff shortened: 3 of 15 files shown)


src/components/atoms/Pre/Pre.styled.tsx
```diff
@@ -1,7 +1,7 @@
 import styled from 'styled-components';
 
 export const StyledPre = styled.pre`
-  padding: 12px 40px 12px 12px;
+  padding: 12px 0 12px 12px;
```
rangoo94, line 4: Wouldn't be better to keep the `padding` both on `left` and `right` side?

```suggestion
  padding: 12px;
```

src/components/custom-antd/Form/Form.styled.tsx
```diff
@@ -1,8 +1,18 @@
...
+const {Item: AntdFormItem} = Form;
+
 export const StyledFormItemLabel = styled.label`
   font-family: ${Fonts.nunito};
   font-weight: 300;
 `;
+
+export const StyledFormItem = styled(AntdFormItem)`
+  flex: 1;
+
```
rangoo94, line 16: Are you sure that it should be applied to all `FormItem`s? Additionally, most likely it means that some existing rules like `width: 100%` should not be needed.

Toxer10: I checked and did not find any `FormItem`s with `width: 100%` CSS rule. Did you mean this?

src/components/atoms/Pre/Pre.styled.tsx
rangoo94: On the screenshot it looks like it's not redundant - there is no space between a button and end of text

src/components/custom-antd/Form/Form.styled.tsx
rangoo94: I meant `FormItem` items that sometimes were enforcing width, because keeping `FormItem`s width was to small for it, although I may be wrong anyway 👍

src/components/atoms/Pre/Pre.styled.tsx
Toxer10: We already have padding on the button to the right. IMO it's totally sufficient to have padding only there. Ofc it's not a big deal to add this padding there.

src/components/custom-antd/Form/Form.styled.tsx
Toxer10: Again, I did not see `FormItem`s with enforced width.

rangoo94: Again, I meant `FormItem` children, not `FormItem` itself :)

haneabogdan (approved):

src/components/atoms/Pre/Pre.styled.tsx
rangoo94: It's padding inside the button, not the margin outside. I'm not a designer though, so probably better to ask @fivenp :)

PR #685, opened 2023-05-17
Move error reports into the area where they were triggered

pavloburchak:
This PR...

## Changes
https://github.com/kubeshop/testkube/issues/2708
- moved error notifications triggered in forms such as creation and settings
- refactored Configuration card, adding ability for future refactor https://github.com/kubeshop/testkube/issues/3787

## Fixes

-

## How to test it
- create test/suite/executor etc
- try to make some error so notification appear
- edit some settings cards across the dashboard

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
+++ b/src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
@@ -1,11 +1,14 @@
-import React, {ReactElement} from 'react';
+import React, {ReactElement, useState} from 'react';
 
 import {Form} from 'antd';
+
+import {ErrorNotificationConfig} from '@models/notifications';
 
 import {Button, Text} from '@custom-antd';
 
 import Colors from '@styles/Colors';
 
+import {NotificationContent} from '../Notification';
 import {
   StyledChildren,
   StyledContainer,
@@ -13,6 +16,7 @@ import {
   StyledFooterButtonsContainer,
   StyledFooterText,
   StyledHeader,
+  StyledNotificationContainer,
 } from './ConfigurationCard.styled';
 
 type ConfigurationCardProps = {
@@ -20,7 +24,7 @@ type ConfigurationCardProps = {
   description: string | ReactElement;
   isWarning?: boolean;
   footerText?: React.ReactNode;
-  onConfirm?: () => void;
+  onConfirm?: () => Promise<ErrorNotificationConfig | null> | void;
   confirmButtonText?: string;
   onCancel?: () => void;
   isButtonsDisabled?: boolean;
@@ -45,6 +49,7 @@ const ConfigurationCard: React.FC<ConfigurationCardProps> = props => {
     isEditable = true,
     enabled = true,
   } = props;
+  const [error, setError] = useState<ErrorNotificationConfig | null>(null);
 
   return (
     <StyledContainer isWarning={isWarning}>
@@ -56,7 +61,16 @@ const ConfigurationCard: React.FC<ConfigurationCardProps> = props => {
           {description}
         </Text>
       </StyledHeader>
-      {children ? <StyledChildren $isActionsVisible={enabled || !isEditable}>{children}</StyledChildren> : null}
+      {children ? (
+        <>
+          {error ? (
+            <StyledNotificationContainer>
+              <NotificationContent status="failed" message={error.message} title={error.title} />
+            </StyledNotificationContainer>
+          ) : null}
+          <StyledChildren $isActionsVisible={enabled || !isEditable}>{children}</StyledChildren>
+        </>
+      ) : null}
       {(enabled && onConfirm) || footerText ? (
         <StyledFooter>
           {footerText && (
@@ -77,10 +91,31 @@ const ConfigurationCard: React.FC<ConfigurationCardProps> = props => {
 
                 return (
                   <StyledFooterButtonsContainer>
-                    <Button onClick={onCancel} $customType="secondary" hidden={!onCancel || disabled}>
+                    <Button
+                      onClick={() => {
+                        onCancel?.();
+                        setError(null);
+                      }}
+                      $customType="secondary"
+                      hidden={!onCancel || disabled}
+                    >
                       Cancel
                     </Button>
-                    <Button onClick={onConfirm} $customType={isWarning ? 'warning' : 'primary'} disabled={disabled} hidden={!onConfirm}>
+                    <Button
+                      onClick={() => {
+                        const confirmResult = onConfirm?.();
+
+                        if (typeof confirmResult === 'object' && typeof confirmResult.then === 'function') {
+                          confirmResult.then((res: any) => {
+                            setError(res);
+                          });
+                        }
+                      }}
+                      $customType={isWarning ? 'warning' : 'primary'}
+                      disabled={disabled}
+                      hidden={!onConfirm}
+                      htmlType="submit"
+                    >
                       {confirmButtonText}
                     </Button>
                   </StyledFooterButtonsContainer>
--- a/src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
+++ b/src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
@@ -108,6 +108,7 @@ const ConfigurationCard: React.FC<ConfigurationCardProps> = props => {
                     </Button>
                     <Button
                       onClick={() => {
+                        setError(null);
                         Promise.resolve(validateFields?.())
                           .then(() => onConfirm?.())
                           .catch((err: ErrorNotificationConfig) => {
--- a/src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
+++ b/src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
@@ -82,7 +82,7 @@ const ConfigurationCard: React.FC<ConfigurationCardProps> = props => {
           )}
           {enabled ? (
             <Form.Item noStyle shouldUpdate>
-              {({isFieldsTouched, getFieldsValue}) => {
+              {({isFieldsTouched, getFieldsValue, validateFields}) => {
                 let disabled = isButtonsDisabled || (getFieldsValue() && !isFieldsTouched());
 
                 if (forceEnableButtons) {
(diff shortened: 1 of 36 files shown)


src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
```diff
@@ -77,10 +91,33 @@ const ConfigurationCard: React.FC<ConfigurationCardProps> = props => {
...
+                      onClick={() => {
+                        validateFields?.().then(() => {
+                          const confirmResult = onConfirm?.();
+
+                          if (typeof confirmResult === 'object' && typeof confirmResult.then === 'function') {
+                            confirmResult.then((res: any) => {
+                              setError(res);
+                            });
+                          }
+                        });
```
rangoo94, line 114: Not 100% sure, but it may be better to make `onConfirm` void, and allow setting `error` as props.

Beside, it rather doesn't make sense to handle `onConfirm`'s result as an error, rather `catch`

src/utils/notification.ts
```diff
@@ -8,21 +9,39 @@ export type DefaultRequestError = {
   status?: number;
 };
 
-export function displayDefaultNotificationFlow<T = unknown>(res?: RTKResponse<T>, success?: () => void) {
-  if (res && 'error' in res && 'data' in res.error) {
-    if (res.error.data) {
+export function defaultNotificationFlow<T = unknown>(
```
rangoo94, line 12: * Looks like it will have a lot of different responsibilities, instead of simply choosing flow. Probably better to extract `getErrorFromResponse` or something like that instead
* Worth to note, that this function is used by Cloud version too, so I would probably avoid changing the name
* Given that this function is synchronous and can't be async - does it make sense to have the callbacks?

src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
```diff
@@ -56,7 +61,16 @@ const ConfigurationCard: React.FC<ConfigurationCardProps> = props => {
           {description}
         </Text>
       </StyledHeader>
-      {children ? <StyledChildren $isActionsVisible={enabled || !isEditable}>{children}</StyledChildren> : null}
+      {children ? (
+        <>
```
rangoo94, line 65: When there are no notifications, there should be at least autoscroll to the error message. Otherwise, the user will not know that something went wrong.

src/components/molecules/Notification/Notification.tsx
```diff
@@ -1,36 +1,12 @@
...
-          <Text className="regular middle">{title}</Text>
-        </StyledNotificationHeader>
-        {message ? (
-          <StyledNotificationMessage>
-            <Text className="regular middle">{message}</Text>
-          </StyledNotificationMessage>
-        ) : null}
-      </StyledNotificationContainer>
-    ),
+    description: <NotificationContent status={status} title={title} message={message} />,
```
rangoo94, line 9: Why the toast notifications should look differently?

rangoo94 (commented):
Reviewed it shortly.

* @fivenp, I'm not convinced to the idea of having these instead of notifications - it may lead to many edge cases (how to react to changes in form - should it be deleted then or after submission, how to ensure the user has seen this message, how to connect it to specific fields or not)
* I feel like `ConfigurationCard` and `displayDefaultNotificationFlow` gets here too much responsibility (from already too big one, especially in `ConfigurationCard`)

src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
pavloburchak: I also was thinking about passing `error` as props, but then I will have to add `useState`s to every component using `ConfigurationCard`. So I think this approach is much better. Also helps to refactor `ConfigurationCard` later.

Sure, I can move handing to `catch`

src/utils/notification.ts
pavloburchak: You mean create another function `getErrorFromResponse` and use it in such cases?

src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
pavloburchak: Ok, I will add it

src/components/molecules/Notification/Notification.tsx
pavloburchak: It is the same, I am just reusing component

src/utils/notification.ts
pavloburchak: Refactored error handing. Looks better now, I think. Do you like it? Or should I move `success` callback to `then` as well?

src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
pavloburchak: Added

src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
```diff
@@ -77,10 +96,35 @@ const ConfigurationCard: React.FC<ConfigurationCardProps> = props => {
...
+                    >
                       Cancel
                     </Button>
-                    <Button onClick={onConfirm} $customType={isWarning ? 'warning' : 'primary'} disabled={disabled} hidden={!onConfirm}>
+                    <Button
+                      onClick={() => {
+                        validateFields?.()
+                          .then(() => {
+                            return onConfirm?.();
+                          })
```
rangoo94, line 114: Optionally:
```suggestion
                          .then(() => onConfirm?.())
```

src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
```diff
@@ -77,10 +96,35 @@ const ConfigurationCard: React.FC<ConfigurationCardProps> = props => {
...
+                      }}
+                      $customType="secondary"
+                      hidden={!onCancel || disabled}
+                    >
                       Cancel
                     </Button>
-                    <Button onClick={onConfirm} $customType={isWarning ? 'warning' : 'primary'} disabled={disabled} hidden={!onConfirm}>
+                    <Button
+                      onClick={() => {
+                        validateFields?.()
```
rangoo94, line 111: When there is no `validateFields` - shouldn't it go through `onConfirm` anyway?
```suggestion
                        Promise.resolve(validateFields?.())
```

src/components/organisms/EntityDetails/EntityDetailsContent/EntityDetailsContent.tsx
```diff
@@ -97,14 +97,18 @@ const EntityDetailsContent: React.FC = () => {
...
         },
       },
-    }).then(res => {
-      displayDefaultNotificationFlow(res, () => {
-        analyticsTrack('trackEvents', {
-          type,
-          uiEvent: `run-${entity}`,
+    })
+      .then(res => {
+        displayDefaultNotificationFlow(res, () => {
```
rangoo94, line 102:
```suggestion
        return displayDefaultNotificationFlow(res, () => {
```
and same for `runEntity`

src/utils/notification.ts
```diff
@@ -8,21 +8,42 @@ export type DefaultRequestError = {
...
-    success && success();
+
+    return errorObject;
   }
+};
+
+export function displayDefaultNotificationFlow<T = unknown>(
+  res?: RTKResponse<T>,
+  success?: () => void
+): {title: string; message?: string} | void {
```
rangoo94, line 39: It's always `void`

src/utils/notification.ts
```diff
@@ -8,21 +8,42 @@ export type DefaultRequestError = {
...
-  } else {
-    success && success();
+
+    return errorObject;
   }
+};
+
+export function displayDefaultNotificationFlow<T = unknown>(
+  res?: RTKResponse<T>,
+  success?: () => void
```
rangoo94, line 38: Maybe instead of `success`, it's better to just keep this function `async` (either by `async` keyword, or by returning `Promise.resolve`/`Promise.reject`).

This way it would be covered with `then`/`catch` flow easily

rangoo94 (commented):
I'm thinking - is it possible and worth considering calling `form.submit`, to keep consistent flow?

src/utils/notification.ts
rangoo94: I think that it would be better to move it to `then` as well, to keep single mechanism :)

pavloburchak: Got it, so we might want to change cloud as well

src/utils/notification.ts
pavloburchak: Yes, sure. Should I change it on cloud as well?

rangoo94: I think so, it will be better to keep it consistent 👍

fivenp:
> @fivenp, I'm not convinced to the idea of having these instead of notifications - it may lead to many edge cases (how to react to changes in form - should it be deleted then or after submission, how to ensure the user has seen this message, how to connect it to specific fields or not)

I'd really like us to give it a try for exactly these reasons..... (e.g. we don't know if they did read the error msg, etc) 

* We know which "settings card" triggered the submit – so we know where to show the error. 
* We would show the error as long as the users starts to re-submit that card OR cancels the changes

rangoo94:
> We would show the error as long as the users starts to re-submit that card OR cancels the changes

Good point, @pavloburchak, could you cover that?

src/components/molecules/ConfigurationCard/ConfigurationCard.tsx
```diff
@@ -77,10 +96,35 @@ const ConfigurationCard: React.FC<ConfigurationCardProps> = props => {
...
+                      }}
+                      $customType="secondary"
+                      hidden={!onCancel || disabled}
+                    >
                       Cancel
                     </Button>
-                    <Button onClick={onConfirm} $customType={isWarning ? 'warning' : 'primary'} disabled={disabled} hidden={!onConfirm}>
+                    <Button
+                      onClick={() => {
+                        Promise.resolve(validateFields?.())
```
rangoo94, line 111: Probably:
```suggestion
                        setError(null);
                        Promise.resolve(validateFields?.())
```

rangoo94 (approved):

PR #688, opened 2023-05-22
tech: read PostHog, Segment and GA4 keys from env variables

rangoo94:
## Changes

- read keys from env variables at build-time

## Fixes

- kubeshop/testkube#3875

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


.github/workflows/docker-build-beta-tag.yml
```diff
@@ -48,7 +48,9 @@ jobs:
           build-args: |
             TARGET=${{ matrix.path }}
             REACT_APP_GOOGLE_ANALYTICS_ID=${{ secrets.REACT_APP_GOOGLE_ANALYTICS_ID }}
-            REACT_APP_SEGMENT_WRITE_KEY=${{ secrets.REACT_APP_SEGMENT_WRITE_KEY }}
+            REACT_APP_SEGMENT_KEY=${{ secrets.REACT_APP_SEGMENT_WRITE_KEY }}
+            REACT_APP_POSTHOG_KEY=phc_DjQgd6iqP8qrhQN6fjkuGeTIk004coiDRmIdbZLRooo
+            REACT_APP_GOOGLE_ANALYTICS_ID=G-945BK09GDC
```
rangoo94, line 53: @ypoplavs, could you add secrets for PostHog and GA ID?

Dockerfile
```diff
@@ -5,8 +5,12 @@ WORKDIR /app
...
 ARG REACT_APP_GOOGLE_ANALYTICS_ID
 ENV REACT_APP_GOOGLE_ANALYTICS_ID=$REACT_APP_GOOGLE_ANALYTICS_ID
-ARG REACT_APP_SEGMENT_WRITE_KEY
-ENV REACT_APP_SEGMENT_WRITE_KEY=$REACT_APP_SEGMENT_WRITE_KEY
+ARG REACT_APP_SEGMENT_KEY
+ENV REACT_APP_SEGMENT_KEY=$REACT_APP_SEGMENT_KEY
+ARG REACT_APP_POSTHOG_KEY
+ENV REACT_APP_POSTHOG_KEY=$REACT_APP_POSTHOG_KEY
+ARG REACT_APP_GOOGLE_ANALYTICS_ID
+ENV REACT_APP_GOOGLE_ANALYTICS_ID=$REACT_APP_GOOGLE_ANALYTICS_ID
```
rangoo94, line 13: @ypoplavs, I recall that in Docker `ARG` is automatically set as `ENV` too - am I right and it could be simplified?

```suggestion
ARG REACT_APP_SEGMENT_KEY
ARG REACT_APP_POSTHOG_KEY
ARG REACT_APP_GOOGLE_ANALYTICS_ID
```

ypoplavs: `ARG` vars are only available during the build stage and **not** in the running container, therefore, we need to have `ENV` for this.

rangoo94: Thank you, so looks like we don't need it, as we set up `ARG` here only for the environment variables that are used in the build-time.

I will keep it the same for now, but have created follow-up ticket to clean it - https://github.com/kubeshop/testkube/issues/3882

PR #697, opened 2023-05-25
feat: adjusted FullWidthSpace component, added new prop

Toxer10:


haneabogdan (approved):

src/styles/Colors.ts
```diff
@@ -69,6 +69,7 @@ enum Colors {
   amber500 = '#F59E0B',
   amber600 = '#D97706',
   amber900 = '#78350F',
+  amber90099 = '#78350F99',
```
rangoo94, line 72: Is `'#78350F99` supported? Shouldn't it be RGBA?

rangoo94: Ahh, actually it should be just deleted as it's not used :)

src/components/custom-antd/FullWidthSpace/FullWidthSpace.styled.tsx
```diff
@@ -1,7 +1,13 @@
...
+import {Space as AntdSpace, SpaceProps} from 'antd';
 
 import styled from 'styled-components';
 
-export const StyledSpace = styled(AntdSpace)`
+export interface FullWidthSpaceProps extends SpaceProps {
+  justify?: 'flex-start' | 'flex-end' | 'center' | 'space-around' | 'space-between';
+}
+
+export const StyledSpace = styled(AntdSpace)<{justify: FullWidthSpaceProps['justify']}>`
```
rangoo94, line 9: Probably `$justify`, to ensure it will not be passed to `AntdSpace`

src/styles/Colors.ts
Toxer10: It's used in Cloud and used directly as is
image

src/components/custom-antd/FullWidthSpace/FullWidthSpace.styled.tsx
Toxer10: Agree

PR #700, opened 2023-05-28
Triggers multi-view layout

pavloburchak:
This PR...

## Changes
- implemented triggers multi view layout
- connected triggers to `zustand` store

https://github.com/kubeshop/testkube/issues/3727
https://github.com/kubeshop/testkube/issues/3765

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/atoms/CreatableMultiSelect/CreatableMultiSelect.tsx
```diff
@@ -62,6 +64,12 @@ const CreatableMultiSelect: React.FC<MultiSelectProps> = props => {
     }
   };
 
+  useEffect(() => {
+    setTimeout(() => {
+      setApplyValidationStyles(validation);
+    }, 100);
+  }, [validation]);
```
rangoo94, line 71:
```suggestion
  import {useDebounce} from 'react-use';
  // ...
  useDebounce(() => {
    setApplyValidationStyles(validation);
  }, 100, [validation]);
```

src/components/atoms/CreatableMultiSelect/CustomComponents/LabelsOption.tsx
```diff
@@ -15,16 +15,9 @@ const LabelsOption = (props: OptionProps<Option>) => {
...
 
-  const allowClick =
-    isChildren && isCreateOption
-      ? Boolean(
-          children.match(
-            /(Create (([A-Za-z0-9][-A-Za-z0-9_./]*)?[A-Za-z0-9]:\s?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?))/g
-          )
-        )
-      : true;
+  const allowClick = isCreateOption ? children.includes(value) : true;
```
rangoo94, line 20: How does it work, or how it was working? This behavior looks *very* strange.

Besides, `allowClick` may be simplified:
```ts
  const allowClick = !isCreateOption || children.includes(value);
```

src/components/molecules/LabelsSelect/LabelsSelect.tsx
```diff
@@ -17,7 +17,7 @@ import {composeLabels} from './utils';
 
 type LabelsSelectProps = {
   onChange?: (value: readonly Option[]) => void;
-  defaultLabels?: readonly Option[];
+  defaultLabels?: Record<string, string> | Option[];
```
rangoo94, line 20: It would be better to keep a single format. Do we actually want `Option[]`? Looking at `composeLabels`, it is actually expecting `Record<any, Option>`.

src/components/organisms/TriggersFormItems/ResourceTriggerSelect/ResourceTriggerSelect.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const {isClusterAvailable} = useContext(MainContext);
+
+  const executors = useAppSelector(selectExecutors);
+
+  const {data: tests = []} = useGetAllTestsQuery(null, {skip: !isClusterAvailable});
+  const {data: testSuites = []} = useGetAllTestSuitesQuery(null, {
+    skip: !isClusterAvailable,
+  });
+
+  const testsData: TestForTrigger[] = useMemo(() => {
```
rangoo94, line 41: It should be resolved easily by `tests` type
```suggestion
  const testsData = useMemo(() => {
```

src/components/organisms/TriggersFormItems/ResourceTriggerSelect/ResourceTriggerSelect.tsx
```diff
@@ -0,0 +1,88 @@
...
+    return tests.map(item => ({
+      name: item.test.name,
+      namespace: item.test.namespace,
+      type: item.test.type,
+    }));
+  }, [tests]);
+
+  const testSuitesData: TestSuiteForTrigger[] = useMemo(() => {
```
rangoo94, line 49: As above
```suggestion
  const testSuitesData = useMemo(() => {
```

src/components/pages/Triggers/TriggerDetails/TriggerDetails.tsx
```diff
@@ -0,0 +1,71 @@
...
+    if (triggerDetails) {
+      setCurrentTrigger(triggerDetails);
+    }
+  }, [triggerDetails]);
+
+  useEffect(() => {
+    if (triggersKeyMap) {
+      setTriggersKeyMap(triggersKeyMap);
+    }
+  }, [triggersKeyMap]);
```
rangoo94, line 49: It's better to use `triggerDetails` and `triggersKeyMap` directly. Otherwise, we will end up having race conditions when switching between different triggers.

src/store/GlobalStore.ts
```diff
@@ -0,0 +1,20 @@
+import {create} from 'zustand';
+import {devtools} from 'zustand/middleware';
+import {shallow} from 'zustand/shallow';
+
+import createTriggersSlice, {TriggersSlice} from './TriggersSlice';
+
+/* Create store. For more information,
+check https://github.com/pmndrs/zustand/blob/main/docs/guides/typescript.md#basic-usage  */
+const useGlobalStore = create<TriggersSlice>()(
```
rangoo94, line 9: It can't be global store, it has to be local. Otherwise, we will i.e. have problems with running it in Cloud.

src/components/pages/Triggers/TriggersList/AddTriggersModal/AddTriggerModal.tsx
```diff
@@ -0,0 +1,179 @@
...
+          <Text className="regular middle">Name</Text>
+          <Input
+            placeholder="e.g.: container-deployment-xyz"
+            value={name}
+            onChange={event => {
+              setName(event.target.value);
+            }}
+          />
+        </Space>
+        <Steps current={currentStep} items={[{title: 'Condition'}, {title: 'Action'}]} />
```
rangoo94, line 114: Is Antd step what we want, or should we move steps from Cloud implementation?

src/components/pages/Triggers/TriggersList/AddTriggersModal/AddTriggerModal.tsx
```diff
@@ -0,0 +1,179 @@
...
+                  marginBottom: 0,
+                }}
+                shouldUpdate
+              >
+                {({getFieldsValue, validateFields}) => (
+                  <Button
+                    $customType="primary"
+                    onClick={() => {
+                      validateFields().then(() => {
+                        setCurrentStep(1);
```
rangoo94, line 136: If `0` and `1` are constants here, it should be rather `enum` or `const`

src/components/pages/Triggers/TriggersList/AddTriggersModal/AddTriggerModal.tsx
```diff
@@ -0,0 +1,179 @@
...
+            placeholder="e.g.: container-deployment-xyz"
+            value={name}
+            onChange={event => {
+              setName(event.target.value);
+            }}
+          />
+        </Space>
+        <Steps current={currentStep} items={[{title: 'Condition'}, {title: 'Action'}]} />
+        <Form layout="vertical" onFinish={onFinish} form={form} name="add-trigger-form">
+          {currentStep === 0 ? (
```
rangoo94, line 116: The steps could be split into separate components.

src/components/pages/Triggers/TriggersList/TriggersList.tsx
```diff
@@ -0,0 +1,122 @@
...
+    if (triggers) {
+      setTriggersList(triggers);
+    }
+  }, [triggers]);
+
+  useEffect(() => {
+    if (triggersKeyMap) {
+      setTriggersKeyMap(triggersKeyMap);
+    }
+  }, [triggersKeyMap]);
```
rangoo94, line 54: It should use `triggers` and `triggersKeyMap` instead

src/components/pages/Triggers/TriggersList/TriggersList.tsx
```diff
@@ -0,0 +1,122 @@
...
+import AddTriggerModal from './AddTriggersModal';
+import EmptyTriggers from './EmptyTriggers';
+import {StyledTriggersGrid, StyledTriggersSkeletonWrapper, TriggerContainer} from './TriggersList.styled';
+
+const Triggers: React.FC = () => {
+  const {triggersList, setTriggersList, setTriggersKeyMap} = useShallowGlobalStore(state => ({
+    triggersList: state.triggersList,
+    setTriggersList: state.setTriggersList,
+    setTriggersKeyMap: state.setTriggersKeyMap,
+  }));
```
rangoo94, line 29: Why do we need it? We use RTK here, and it's not used anywhere else in app

src/styles/globalStyles.ts
```diff
@@ -374,6 +374,49 @@ export const GlobalStyle = createGlobalStyle`
...
     letter-spacing: 0.05em;
   }
+
+  /* Steps */
+
+  .ant-steps-item-active {
+    .ant-steps-item-container {
+      .ant-steps-item-content {
+        .ant-steps-item-title {
+          color: ${Colors.indigo400};
```
rangoo94, line 384: Isn't it controlled by props or LESS? Similar for below.

rangoo94:
Can we split this PR to smaller ones? This PR is meant to change the display of a single page, yet it touches 50+ files with many changes from different contexts.

I don't think it would be possible to review this thoroughly, as it's too complex a mix of changes - it may require full focus for a few hours, just to understand the connections between different changes.

src/styles/globalStyles.ts
pavloburchak: It is not. Maybe I am missing something, you can check it here https://github.com/ant-design/ant-design/blob/4.x-stable/components/style/themes/default.less
I always try to resolve it with Less, if it is possible

src/components/pages/Triggers/TriggersList/TriggersList.tsx
pavloburchak: What exactly and why?

rangoo94: Basically,

* RTK returns now data to `triggers` / `triggersKeyMap` variables,
* These effects are moving these values (when not falsy) into `triggersList` / `triggersKeyMap`
* The whole code here is using `triggersList` / `triggersKeyMap` is unused

Instead, we could use here data directly from RTK, as these are recent.

rangoo94: Ahh, actually `triggersKeyMap` is used, just in different places; additionally, it's downloaded and set also from multiple places though. I'm not sure what should be there, it will require much more complex analysis than what I did for now.

src/components/atoms/CreatableMultiSelect/CustomComponents/LabelsOption.tsx
pavloburchak: It is basically allowing to click on Create option only is value passes validation. It is achieved by formatting `formatCreateLabel` prop in `LabelsSelect` component. And here I check if this option includes the value, because in `formatCreateLabel` in add it when validation passes

rangoo94: So we add some *magic* options like `Create Something` to the list of options, and then this option has some action? It sounds like this mechanism is unnecessarily obscure, instead of *actually* passing options for creation...

I'm not sure if it's worth adjusting it here with another hack - maybe it's better to keep it as it is, and just refactor soon.

src/components/pages/Triggers/TriggersList/AddTriggersModal/AddTriggerModal.tsx
pavloburchak: Yes, it is exactly what we want

src/components/pages/Triggers/TriggersList/AddTriggersModal/AddTriggerModal.tsx
pavloburchak: Ok, i will change it to enum

src/components/pages/Triggers/TriggersList/TriggersList.tsx
pavloburchak: I tried to repeat pattern as we have for other entities. I can remove it from store if you think we should not keep it just in case

src/components/pages/Triggers/TriggersList/TriggersList.tsx
pavloburchak: It is basically options for triggers. User can come to list or details page and should have them available. Basically it is possible to fetch them as we fetch executors. But the difference here is that `triggersKeyMap` we use only in triggers, but executors could be present anywhere

src/components/atoms/CreatableMultiSelect/CustomComponents/LabelsOption.tsx
rangoo94: Based on https://react-select.com/props#creatable-props it should be simple:
* it may be formatted with `formatCreateLabel` prop
* it may be detected with `onCreateOption` callback

src/components/organisms/TriggersFormItems/ActionFormItems.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <Select options={actionOptions} placeholder="Select a testkube related action" />
+      </Form.Item>
+      <Form.Item noStyle shouldUpdate>
+        {({getFieldValue}) => {
+          const label = getFieldValue('testLabelSelector');
+          const name = getFieldValue('testNameSelector');
+
+          if (isFirst) {
+            setSwitcherValue(name ? 'name' : 'label');
+          }
```
rangoo94, line 44: * Calling `setState` during render causes errors
* Should it really happen on the first mount only? If so, probably `useEffect(() => {}, [])` could be used

src/components/organisms/TriggersFormItems/ActionFormItems.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import ResourceTriggerSelect from './ResourceTriggerSelect';
+import TriggerSelectorSwitcher from './TriggerSelectorSwitcher';
+
+const ActionFormItems = () => {
+  const {triggersKeyMap} = useStore(state => ({
+    triggersKeyMap: state.triggersKeyMap!,
+  }));
+
+  const [switcherValue, setSwitcherValue] = useState<string | number>('label');
```
rangoo94, line 20: Based on the code, it's rather always `string`, or even more `'name' | 'label'`. Could be extracted to literal anyway.

src/components/organisms/TriggersFormItems/ActionFormItems.tsx
```diff
@@ -0,0 +1,64 @@
...
+          return (
+            <Space size={16} direction="vertical" style={{width: '100%'}}>
+              <TriggerSelectorSwitcher value={switcherValue} onChange={setSwitcherValue} />
+              {switcherValue === 'label' ? (
+                <Form.Item label="Testkube resource" required name="testLabelSelector" rules={[required]}>
+                  <LabelsSelect defaultLabels={label} />
+                </Form.Item>
+              ) : (
+                <ResourceTriggerSelect />
+              )}
```
rangoo94, line 55: Wouldn't be easier to extract it to separate component? No `switcherValue` logic will need to be there.

Besides, `ResourceTriggerSelect` could support input properties, so then there is no `Form.Item` included inside.

src/components/organisms/TriggersFormItems/ConditionFormItems.tsx
```diff
@@ -0,0 +1,79 @@
...
+                  <Input placeholder="e.g.: namespace/resource-name" />
+                </Form.Item>
+              )}
+            </Space>
+          );
+        }}
+      </Form.Item>
+      <Form.Item noStyle shouldUpdate>
+        {({getFieldValue}) => {
+          let eventsOptions;
```
rangoo94, line 60: Why not constant? Looks like it's assigned just once.

src/components/pages/Triggers/TriggerDetails/TriggerDetails.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  useEffect(() => {
+    safeRefetch(refetch);
+    safeRefetch(refetchKeyMap);
+  }, [location]);
```
rangoo94, line 53: That has no effect (checked that). Besides:

* `triggersKeyMap` rather doesn't have to be refetched, or is it dynamic?
*  instead of `location` it could base on `name`

src/components/pages/Triggers/TriggerDetails/TriggerDetails.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const {isClusterAvailable} = useContext(MainContext);
+  const {location, navigate} = useContext(DashboardContext);
+  const {pageTitle} = useContext(ConfigContext);
+
+  const {setCurrentTrigger, setTriggersKeyMap} = useStore(state => ({
+    setCurrentTrigger: state.setCurrentTrigger,
+    setTriggersKeyMap: state.setTriggersKeyMap,
+  }));
+
+  const name = useLocation().lastPathSegment;
```
rangoo94, line 29:
```suggestion
  const name = useParams().id!;
```

src/components/pages/Triggers/TriggersList/TriggersList.tsx
```diff
@@ -0,0 +1,114 @@
...
+const Triggers: React.FC = () => {
+  const {setTriggersKeyMap} = useStore(state => ({
+    setTriggersKeyMap: state.setTriggersKeyMap,
+  }));
+
+  const {isClusterAvailable} = useContext(MainContext);
+  const {location, navigate} = useContext(DashboardContext);
+
+  const {data: triggersList = [], refetch, isLoading} = useGetTriggersListQuery(null, {skip: !isClusterAvailable});
+  const {data: triggersKeyMap, refetch: refetchKeyMap} = useGetTriggersKeyMapQuery(null, {skip: !isClusterAvailable});
```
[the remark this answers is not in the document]
pavloburchak: Probably no, it is needed in AddTrigger modal, so I might move it there

src/components/organisms/TriggersFormItems/ActionFormItems.tsx
pavloburchak: - What kind of errors in this case?
- Do you think it is clear to extract `name` and pass it as props so I can check it is `useEffect`? It adds too much complexity I think

src/components/organisms/TriggersFormItems/ActionFormItems.tsx
pavloburchak: What exactly should be extracted? `switcherValue` logic is still going to be somewhere, I don't understand downside of this approach

src/components/organisms/TriggersFormItems/ResourceTriggerSelect/ResourceTriggerSelect.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const testSuitesData = useMemo(() => {
+    return testSuites.map(item => ({
+      name: item.testSuite.name,
+      namespace: item.testSuite.namespace,
+    }));
+  }, [testSuites]);
+
+  return (
+    <Form.Item label="Testkube resource" required name="testNameSelector" rules={[required]}>
```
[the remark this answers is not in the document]
pavloburchak: It can't be. I would like it to be just `Select` too, but without `Form.Item` in this component, it doesn't really work. Default value not provided to select, as well as `isFieldsTouched` not triggered when I select something else.

src/components/organisms/TriggersFormItems/ActionFormItems.tsx
rangoo94: React internal errors, like:
```
Cannot update during an existing state transition (such as within render or another component's constructor). Render methods should be a pure function of props and state.
```

src/components/organisms/TriggersFormItems/ResourceTriggerSelect/ResourceTriggerSelect.tsx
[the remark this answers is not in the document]
rangoo94: * The default value doesn't have to be provided - there is no value, then it's default. Additionally, you may pass any custom props you want to the input.
* `isFieldsTouched` will be updated when you update the value. When you handle the `onChange` prop in your custom input, any call to it will just make the field touched and the value updated.

src/components/organisms/TriggersFormItems/ActionFormItems.tsx
rangoo94: On the other side, when digging more in the code, I think that the whole idea of `ActionFormItems` has a flaw - it's treated as a kind-of Form trait, where multiple fields could be treated out of the context.

In this case, we may just ignore it, and refactor it later - it may be too complex to fix it now.

src/components/organisms/TriggersFormItems/TriggerSelectorSwitcher/TriggerSelectorSwitcher.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const {value, onChange} = props;
+
+  return (
+    <Segmented
+      options={[
+        {label: 'BY LABEL', value: 'label'},
+        {label: 'BY NAME', value: 'name'},
+      ]}
+      value={value}
+      onChange={x => onChange(String(x))}
```
rangoo94, line 18: Is there needed the casting to `String`? Looks like there needs to be some different logic then

src/components/pages/Triggers/TriggersList/AddTriggerModal/ModalSecondStep.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  return (
+    <>
+      <StyledStepDescription>
+        <Text color={Colors.slate400} className="regular middle">
+          Define the action to be performed on testkube once the conditions are met.
+        </Text>
+      </StyledStepDescription>
+      <ActionFormItems />
+      <Form.Item style={{marginBottom: 0}} shouldUpdate>
```
rangoo94, line 28: Would `Formitem` from `@custom-antd` be better?

src/components/pages/Triggers/TriggersList/AddTriggerModal/ModalSecondStep.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <StyledButtonsContainer>
+            <Button
+              $customType="secondary"
+              onClick={() => {
+                setCurrentStep(StepsEnum.condition);
+              }}
+            >
+              Back
+            </Button>
+            <Button htmlType="submit" $customType="primary" loading={isLoading} disabled={!isFieldsTouched()}>
```
rangoo94, line 39:
```suggestion
            <Button htmlType="submit" $customType="primary" loading={isLoading} disabled={!isFieldsTouched() || isLoading}>
```

src/components/pages/Triggers/TriggersList/TriggersList.tsx
```diff
@@ -0,0 +1,100 @@
...
+import {useGetTriggersListQuery} from '@services/triggers';
+
+import {externalLinks} from '@utils/externalLinks';
+import {safeRefetch} from '@utils/fetchUtils';
+
+import AddTriggerModal from './AddTriggerModal';
+import EmptyTriggers from './EmptyTriggers';
+import {StyledTriggersGrid, StyledTriggersSkeletonWrapper, TriggerContainer} from './TriggersList.styled';
+
+const Triggers: React.FC = () => {
```
rangoo94, line 22:
```suggestion
const TriggersList: React.FC = () => {
```

src/components/pages/Triggers/TriggersList/TriggersList.tsx
```diff
@@ -0,0 +1,100 @@
...
+          setIsModalVisible={setAddTriggerModalVisibility}
+          width={546}
+          content={<AddTriggerModal />}
+        />
+      ) : null}
+    </PageBlueprint>
+  );
+};
+
+export default Triggers;
```
rangoo94, line 100:
```suggestion
export default TriggersList;
```

src/components/pages/Triggers/utils.ts
```diff
@@ -1,24 +1,36 @@
...
-  },
-];
+import {decomposeLabels} from '@molecules/LabelsSelect/utils';
+
+export const getResourceIdentifierSelector = (
+  formValue: string | readonly Option[],
+  appNamespace: string
+): TestTriggerSelector => {
+  if (typeof formValue === 'string') {
+    if (formValue?.includes('/')) {
```
rangoo94, line 11: There is already known that the `formValue` is `string`:
```suggestion
    if (formValue.includes('/')) {
```

src/components/organisms/TriggersFormItems/ActionFormItems.tsx
rangoo94: Marking as resolved then, and we will just handle it later 👍

src/components/organisms/TriggersFormItems/ActionFormItems.tsx
rangoo94: As in a comment above, marking as resolved for now 👍

src/components/pages/Triggers/TriggerDetails/TriggerSettings/General/Name.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <ConfigurationCard
+        title="Trigger name"
+        description="Define the name of your trigger"
+        onConfirm={onFinish}
+        onCancel={() => {
+          form.resetFields();
+        }}
+        enabled={mayEdit}
+      >
+        <Form.Item label="Name" required name="name">
```
rangoo94, line 56:
```suggestion
        <Form.Item label="Name" name="name">
```

src/components/pages/Triggers/TriggerDetails/TriggerSettings/General/Name.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <Form form={form} name="general-settings-name" initialValues={{name}} layout="vertical" disabled={!mayEdit}>
+      <ConfigurationCard
+        title="Trigger name"
+        description="Define the name of your trigger"
+        onConfirm={onFinish}
+        onCancel={() => {
+          form.resetFields();
+        }}
+        enabled={mayEdit}
```
rangoo94, line 54: The form is not editable, and there is no need for `mayEdit` logic because of that too.

```suggestion
        isButtonsDisabled
        enabled={false}
        isEditable={false}
```

src/components/organisms/TriggersFormItems/ResourceTriggerSelect/ResourceTriggerSelect.tsx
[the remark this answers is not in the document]
pavloburchak: I might not get your point here. Isn't it more simple to leave it as it is? If I remove FormItem I would have to Create custom label, required mark, control the value with `onChange` and `value` props. What is the benefits of this approach, or I didn't get your point correctly?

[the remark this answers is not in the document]
rangoo94: You don't have to add custom labels or required marks. The `Form.Item` should be kept, just a level above - in the form where `ResourceTriggerSelect` is actually used.

The `onChange` and `value` will need to be controlled though, but it's just a few lines of a common code and will make it *much* simpler for unit tests and maintenance.

src/components/organisms/TriggersFormItems/TriggerSelectorSwitcher/TriggerSelectorSwitcher.tsx
pavloburchak: `Type 'number' is not assignable to type 'string'.`
I get this if I remove it. Originally yes, it has to be string or number, but I removed it as you asked

rangoo94: Ok, it's fine then 👍

src/components/pages/Triggers/TriggersList/AddTriggerModal/AddTriggerModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+      <div ref={topRef} />
+      <Space size={24} direction="vertical" style={{width: '100%', marginTop: 0}}>
+        <Space direction="vertical" style={{width: '100%'}}>
+          {error ? (
+            <StyledNotificationContainer>
+              <NotificationContent status="failed" message={error.message} title={error.title} />
+            </StyledNotificationContainer>
+          ) : null}
+          <Text className="regular middle">Name</Text>
+          <Input
```
rangoo94, line 122: It should be required and probably just should be a part of the Form below

src/components/organisms/TriggersFormItems/ActionFormItems.tsx
```diff
@@ -0,0 +1,66 @@
...
+          const name = getFieldValue('testNameSelector');
+
+          if (isFirst) {
+            setSwitcherValue(name ? 'name' : 'label');
+          }
+
+          return (
+            <Space size={16} direction="vertical" style={{width: '100%'}}>
+              <TriggerSelectorSwitcher value={switcherValue} onChange={setSwitcherValue} />
+              {switcherValue === 'label' ? (
```
rangoo94, line 49: Bug: Looks like it's keeping both `testLabelSelector` and `testNameSelector` even after saving the data

src/components/pages/Triggers/TriggersList/AddTriggerModal/AddTriggerModal.tsx
rangoo94: Looks like it's fine - it is optional, as it may be autogenerated when empty

src/components/organisms/TriggersFormItems/ConditionFormItems.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const {triggersKeyMap, currentTrigger} = useStore(state => ({
+    triggersKeyMap: state.triggersKeyMap!,
+    currentTrigger: state.currentTrigger!,
+  }));
+
+  const [switcherValue, setSwitcherValue] = useState('label');
+
+  const {
+    resourceSelector: {name: nameSelector},
+  } = currentTrigger;
```
rangoo94, line 23: Throws error on "Create Trigger" modal:

```
ConditionFormItems.tsx:22 Uncaught TypeError: Cannot destructure property 'resourceSelector' of 'currentTrigger' as it is undefined.
    at ConditionFormItems (ConditionFormItems.tsx:22:1)
```
```

Comment (rangoo94) [reply src/components/organisms/TriggersFormItems/ConditionFormItems.tsx:23]:
```suggestion
  const nameSelector = Form.useFormInstance().getFieldValue('resourceNameSelector');
```

rangoo94 (approved):
LGTM, tested and works fine 👍

PR #703, opened 2023-05-29
Implement Triggers store in Zustand

pavloburchak:
This PR...

## Changes
- https://github.com/kubeshop/testkube/issues/3765

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/store/LocalStore.ts
```diff
@@ -0,0 +1,20 @@
+import {create} from 'zustand';
+import {devtools} from 'zustand/middleware';
+import {shallow} from 'zustand/shallow';
+
+import createTriggersSlice, {TriggersSlice} from './TriggersSlice';
+
+/* Create store. For more information,
+check https://github.com/pmndrs/zustand/blob/main/docs/guides/typescript.md#basic-usage  */
+const useLocalStore = create<TriggersSlice>()(
```
rangoo94, line 9: It is a global store, yet it should be local = without global space pollution. @haneabogdan, could you assist with that, based on your findings?

src/store/TriggerLocalStore.ts
```diff
@@ -0,0 +1,25 @@
...
+          )
+    );
+    const useStore = useStoreRef.current;
+          
+    const useShallowLocalStore = <U>(selector: (state: TriggersSlice) => U) => useStore(selector, shallow);
+
+    return [useShallowLocalStore];
+}; 
+
+export default useTriggerLocalStore;
```
haneabogdan, line 25: this is a draft implementation of local store.
Usage would be:

`  const [useShallowLocalStore] = useTriggerLocalStore();
// whatever props we want to fetch from store
  const {text, setText} = useShallowLocalStore(state => ({
    text: state.text,
    setText: state.setText,
  }));`

haneabogdan: Tested it locally with 2 instances of the same store on different components

rangoo94: Nice one! Is there a way to provide the initial state?

pavloburchak: I believe it is provided in TriggersSlice itself, isn't is @haneabogdan ?

haneabogdan (approved):

PR #704, opened 2023-05-29
Allow using Prettier from different Working Directory

pavloburchak:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,6 +1,8 @@
 const {loadConfig: loadTsConfig} = require('tsconfig-paths');
 
-const paths = Object.keys(loadTsConfig('./tsconfig.json').paths)
+const {join} = require('node:path');
+
+const paths = Object.keys(loadTsConfig(join(__dirname, 'tsconfig.json')).paths)
   .map(path => path.replace(/\/\*$/, ''))
   .sort()
   .filter((x, i, a) => a.indexOf(x) === i);


.prettierrc.js
```diff
@@ -1,6 +1,6 @@
 const {loadConfig: loadTsConfig} = require('tsconfig-paths');
 
-const paths = Object.keys(loadTsConfig('./tsconfig.json').paths)
+const paths = Object.keys(loadTsConfig(require('node:path').join(__dirname, 'tsconfig.json')).paths)
```
rangoo94, line 3: `const {join} = require('node:path');` should be on top

src/services/apiEndpoint.spec.ts
```diff
@@ -20,7 +20,7 @@ import {
   useWsEndpoint,
 } from './apiEndpoint';
 
-function replaceEnvEach<K extends keyof typeof env>(name: K, value: (typeof env)[K]): void {
+function replaceEnvEach<K extends keyof typeof env>(name: K, value: typeof env[K]): void {
```
rangoo94, line 23: Was it from Prettier? If so, I need to find a reason why it wasn't detected before.

rangoo94 (approved):
Trivial fix needed, so marking as approved.

The title is misleading though, it's unrelated to "Apply prettier on file save".

* It's not adding a feature of applying Prettier on file save
* It's possible to setup Prettier on file save without this fix
* This PR is only to allow using Prettier from different Working Directory, than root directory here

src/services/apiEndpoint.spec.ts
rangoo94: It wasn't from Prettier, because it has reported a problem in this PR.
```suggestion
function replaceEnvEach<K extends keyof typeof env>(name: K, value: (typeof env)[K]): void {
```

pavloburchak: It was formatted for me after `npm run format:all`. Should I remove it?

pavloburchak:
> Trivial fix needed, so marking as approved.
> 
> The title is misleading though, it's unrelated to "Apply prettier on file save".
> 
> * It's not adding a feature of applying Prettier on file save
> * It's possible to setup Prettier on file save without this fix
> * This PR is only to allow using Prettier from different Working Directory, than root directory here

Got it, thanks

src/services/apiEndpoint.spec.ts
rangoo94: Prettier says that it's invalid, so it should be rather reverted

PR #735, opened 2023-06-09
Split EntityList logic (Tests & Test Suites)

rangoo94:
This PR is trying to extract the abstract logic from the EntityList page, that was handling both Tests and Test Suites.
There may be still another stage of improving that, yet it already has decoupled a lot, reducing its code by half, and works well as such.

## Changes

- **Note: it may be easier to check this PR commit by commit - it has been iteratively decoupling these pages, so it's easier to understand the reasoning this way**
- Prepare specific `TestsList` and `TestSuitesList` pages, instead of the `EntityListBlueprint`
- Avoid all the related configuration files and data layer
- use `ModalContext` instead of local `Modal`
- avoid adding `testIcon` property into the Test - compute it dynamically
- Delete unused or unnecessary code
- Simplify deep-linking logic (filters in the URL) - handle it fully in `EntityListContent`

## Fixes

- part of https://github.com/kubeshop/testkube/issues/3958

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


Toxer10 (approved):

rangoo94:
> Filters is supposed to be remain when you refresh, or paste a link with them applied. I see it has been removed in this PR, is it meant to happen?

It worked fine for the production build, it wasn't working for the `StrictMode` available for development. Fixed now 👍

pavloburchak (approved):

pavloburchak:
image
Abort option looks not really great when the item is on the right

pavloburchak:
Also according abortion, I am not sure it is a great idea to display this option when there are no running executions. I click on it and nothing happens

rangoo94:
> Also according abortion, I am not sure it is a great idea to display this option when there are no running executions. I click on it and nothing happens

@fivenp, what do you think about it?

fivenp:
I think: https://github.com/kubeshop/testkube/issues/3988 :)

PR #746, opened 2023-06-23
feat: Download all artifacts

pavloburchak:
This PR...

## Changes

- download artifacts archive

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/ArtifactsList/ArtifactsList.tsx
+++ b/src/components/molecules/ArtifactsList/ArtifactsList.tsx
@@ -1,13 +1,18 @@
-import {useMemo} from 'react';
+import {useMemo, useState} from 'react';
 
-import {Skeleton, Text} from '@custom-antd';
+import {Button, Skeleton, Text} from '@custom-antd';
 
 import {Artifact} from '@models/artifact';
 
+import {downloadArtifactArchive} from '@services/artifacts';
+
 import Colors from '@styles/Colors';
+
+import {DefaultRequestError, displayDefaultErrorNotification} from '@utils/notification';
 
 import {ArtifactsListContainer} from './ArtifactsList.styled';
 import ArtifactsListItem from './ArtifactsListItem';
+import {StyledDownloadAllContainer} from './ArtifactsListItem.styled';
 
 type ArtifactsListProps = {
   artifacts: Artifact[];
@@ -19,6 +24,8 @@ type ArtifactsListProps = {
 
 const ArtifactsList: React.FC<ArtifactsListProps> = props => {
   const {artifacts, testExecutionId, testName, testSuiteName, isLoading} = props;
+
+  const [isDownloading, setIsDownloading] = useState(false);
 
   const renderedArtifactsList = useMemo(() => {
     if (isLoading) {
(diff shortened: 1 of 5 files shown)


src/services/artifacts.ts
```diff
@@ -39,3 +39,30 @@ export const downloadArtifact = async (
...
+  if (response.headers.get('content-type') === 'application/json') {
+    response = await fetch((await response.json()).data.url);
+  }
+
+  // Download the file
+  const blobUrl = window.URL.createObjectURL(await response.blob());
+  const a = document.createElement('a');
+  a.href = blobUrl;
+  // tar.gz format for archive
+  a.download = `${fileName}.tar.gz`;
```
rangoo94, line 63: Shouldn't the `fileName` have the extension itself?

src/components/molecules/ArtifactsList/ArtifactsList.tsx
```diff
@@ -55,7 +62,27 @@ const ArtifactsList: React.FC<ArtifactsListProps> = props => {
     });
   }, [artifacts, testExecutionId, isLoading]);
 
-  return <ArtifactsListContainer>{renderedArtifactsList}</ArtifactsListContainer>;
+  const handleDownloadAll = async () => {
+    try {
+      setIsDownloading(true);
+      await downloadArtifactArchive(`${testName}-${testExecutionId}`, testExecutionId);
```
rangoo94, line 68: Maybe it's worth to add a date to this name?
```
      await downloadArtifactArchive(`${testName}-${testExecutionIsoDateOrTimestamp}-${testExecutionId}`, testExecutionId);
```

rangoo94 (approved):

src/services/artifacts.ts
pavloburchak: I thought I shouldn't pollute Artifacts list component with it. For example if it would be required to download archive from some other place, we would have to not forget about specifying this format.

src/components/molecules/ArtifactsList/ArtifactsList.tsx
pavloburchak: Maybe, but this file name is still pretty big. What benefits of adding timestamps here?

src/services/artifacts.ts
rangoo94: I don't think that it's pollution, the extension is part of the file name. The archive in different places could be of a different format (in theory, because in practice I doubt so)

src/components/molecules/ArtifactsList/ArtifactsList.tsx
rangoo94: When the User will download artifacts from multiple executions and will sort the files, will get a list from the latest to the oldest

rangoo94: Actually, I think that most often the execution ID is not random ID - it's `test-name-1234`. It's not always happening though, as:
* I've seen sometimes the prefix missing
* The execution name may be overridden

pavloburchak: Ok

src/services/artifacts.ts
pavloburchak: Ok

PR #753, opened 2023-06-28
feat: e2e test logs

haneabogdan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


test/dashboard-e2e/tests/test-logs.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+import {test} from '@playwright/test';
+
+import {ApiHelpers} from '../api/api-helpers';
+import {TestDataHandler} from '../data-handlers/test-data-handlers';
+import {MainPage} from '../pages/MainPage';
+import {TestExecutionsPage} from '../pages/TestExecutionsPage';
+
+const apiHelpers = new ApiHelpers(process.env.API_URL, process.env.CLOUD_CONTEXT, process.env.BEARER_TOKEN);
+
+test(`Run test logs`, async ({page}) => {
```
tkonieczny, line 11: Logs are something that can/should be checked in `run-test.spec.ts` - separate test isn't needed.
There are 3 cases there (`['cypress-git-created', 'k6-git-created', 'postman-git-created']`). All of them are currently aborted not to waste resources, and because cypress execution can be slow. In these case I think it's better to execute k6 or postman test, and check the logs there.

test/dashboard-e2e/pages/TestExecutionsPage.ts
```diff
@@ -1,16 +1,27 @@
...
+
+  async checkWebSocketOpened() {
+    const websocket = await this.page.waitForEvent('websocket');
+    const isClosed = await websocket.isClosed();
+    await expect(isClosed).toBeFalsy();
+  }
+
+  async checkExecutionLogs() {
+    const logs = await this.page.locator('//code');
+    await expect(logs).toContainText(`{"status":"running"}`);
```
tkonieczny, line 25: - `validateExecutionLogContents`? 
- expected text should be passed as argument

Should the logs really contain `{"status":"running"}`?

haneabogdan: - naming suggestion is great

first output of the logs stream is always something like `{"status":"running"}` . I would need to timeout the tests for more than 10-15 seconds to get more output from the stream. maybe by using k6 or postman tests, it will be faster. will check

rangoo94: The `{"status":"running"}` is not necessarily there, for example, it doesn't appear for the finished test, and to be fair it's kind of a bug that it appears at all.

Maybe it would be better to add `async hasExecutionLogs(message: string)` instead, so it will be easier to use?

test/dashboard-e2e/pages/TestExecutionsPage.ts
```diff
@@ -1,16 +1,27 @@
...
+  }
+
+  async checkWebSocketOpened() {
+    const websocket = await this.page.waitForEvent('websocket');
+    const isClosed = await websocket.isClosed();
+    await expect(isClosed).toBeFalsy();
+  }
+
+  async checkExecutionLogs() {
+    const logs = await this.page.locator('//code');
```
rangoo94, line 24: `//code` is a too-wide selector. Maybe add `data-test` to this element?

test/dashboard-e2e/pages/TestExecutionsPage.ts
haneabogdan: agreed. I'll find a better message to assert against. As for the API, let's try to remove this message, since it's not relevant for the user

rangoo94: I would consider as well `message: string | RegExp` - probably to detect logs it would be enough to check for `[^\t]+` 🙂

rangoo94:
@haneabogdan, could you rebase on top of `develop`? Looks like there are some conflicting changes, that may make diff less clear

rangoo94 (approved):

PR #765, opened 2023-07-10
fix: fixed a problem reported on sentry io

Toxer10:
This change should fix [this](https://kubeshop.sentry.io/issues/4297803908/?project=4504718232911872&query=is%3Aunresolved&referrer=issue-stream&stream_index=1).

### DIFF
--- a/src/utils/sources.ts
+++ b/src/utils/sources.ts
@@ -129,7 +129,7 @@ export type GetSourceFormValues = {
 export const getSourceFormValues = (entityDetails: Test, testSources: SourceWithRepository[]): GetSourceFormValues => {
   const {content} = entityDetails;
 
-  if (!content || !content.type) {
+  if (!content?.type) {
     return {source: ''};
   }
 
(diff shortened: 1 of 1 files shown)


src/utils/sources.ts
```diff
@@ -129,6 +129,10 @@ export type GetSourceFormValues = {
 export const getSourceFormValues = (entityDetails: Test, testSources: SourceWithRepository[]): GetSourceFormValues => {
   const {content} = entityDetails;
 
+  if (!content || !content.type) {
```
rangoo94, line 132:
```suggestion
  if (!content?.type) {
```

Toxer10: In general it shouldn't be acceptable but for some reason in under some circumstances it's not present.

rangoo94 (approved):

rangoo94:
I don't think the solution is valid, we should rather check why such test happened and either display error or fix a bug, but I'll keep it on my list to fix some day, good to not have the page fully breaking 👍

PR #775, opened 2023-07-13
tests for strings util

Toxer10:


src/utils/strings.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+describe('strings', () => {
+  describe('hasProtocol', () => {
+    it('should return true if the url has a protocol', () => {
+      expect(hasProtocol('https://www.google.com')).toBe(true);
+      expect(hasProtocol('http://www.google.com')).toBe(true);
+    });
+
+    it('should return false if the url does not have a protocol', () => {
+      expect(hasProtocol('www.google.com')).toBe(false);
+      expect(hasProtocol('google.com')).toBe(false);
```
rangoo94, line 12: We could check cases with incorrectly placed protocols
```suggestion
      expect(hasProtocol('google.com')).toBe(false);
      expect(hasProtocol('http:google.com')).toBe(false);
      expect(hasProtocol('http:/google.com')).toBe(false);
```

src/utils/strings.spec.ts
```diff
@@ -0,0 +1,15 @@
+import {hasProtocol} from './strings';
+
+describe('strings', () => {
+  describe('hasProtocol', () => {
+    it('should return true if the url has a protocol', () => {
+      expect(hasProtocol('https://www.google.com')).toBe(true);
+      expect(hasProtocol('http://www.google.com')).toBe(true);
```
rangoo94, line 7: We could check other protocols too.
```suggestion
      expect(hasProtocol('http://www.google.com')).toBe(true);
      expect(hasProtocol('ftp://www.google.com')).toBe(true);
```

src/utils/strings.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+    });
+
+    it('should return false if the url does not have a protocol', () => {
+      expect(hasProtocol('www.google.com')).toBe(false);
+      expect(hasProtocol('google.com')).toBe(false);
+    });
```
rangoo94, line 13: We could also add a test that would define that we do not accept the propagated protocol (`//`).
```suggestion
    });

    it('should disallow the shared protocol', () => {
      expect(hasProtocol('//')).toBe(false);
    });
```

rangoo94 (approved):

rangoo94:
Additionally, we could check URLs with paths, to ensure that it will not work for domains only

PR #786, opened 2023-07-18
feat: Plugin Manager

haneabogdan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/utils/createPluginManager.ts
```diff
@@ -0,0 +1,39 @@
+import {PluginData, PluginKeys, Plugins} from '@contexts/PluginsContext';
+
+export class PluginManager {
+  private plugins: Plugins;
+
+  public constructor(plugins?: Plugins) {
+    plugins ? (this.plugins = plugins) : (this.plugins = initDefaultPlugins());
+  }
+
+  public get(key: PluginKeys): PluginData[] {
```
rangoo94, line 10: There should be no need to get plugin itself - only the data that it injects

rangoo94 (commented):
The main problem here, is that in this implementation Plugin is not actual plugin - it's just single set of data. Because of that, you cannot add any complex functionality - i.e. adding AI with 2 tabs would mean that you have to create and inject 2 plugins.

Proper solution would be to allow Plugins to actually inject to different (optionally: defined, but better not) places.

As a simple example:

```ts
manager.add({
  name: 'plugin',
  executionDetailsTabs: [
    { label: 'XYZ', content: 'ABC' },
  ],
});
```

Or more flexible:

```ts
manager.add({
  name: 'plugin',
  append: {
    executionDetailsTabs: [
      { label: 'XYZ', content: 'ABC' },
    ],
  },
});
```

As an alternative, Plugins could just have only a behavior - that would make them much more flexible:

```ts
manager.add({
  name: 'plugin',
  setup: (app: App) => {
    app.append('executionDetailsTabs', [
      {label: 'XYZ', content: 'ABC'},
    ]);
  },
```

src/plugins/createPluginManager.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export class PluginManager {
+  private plugins: Plugin[];
+
+  public constructor() {
+    this.plugins = [];
+  }
+
+  public add(plugin: Plugin): void {
+    plugin.order ? this.plugins.splice(plugin.order, 0, plugin) : this.plugins.push(plugin);
```
rangoo94, line 11: Order isn't an index, it's a priority. Imagine in example:

```ts
existingOrders = [ 5, 5, 5, 3, 2, 1 ]
newOrder = 1
newOrders = [ 5, 1, 5, 5, 3, 2, 1 ] // wrong
```

Besides, it's good to have this array immutable, so the changes could be detected.

```suggestion
    const order = plugin.order ?? -Infinity;
    const index = this.plugins.findIndex(item => item.order < order);
    if (index === -1) {
      this.plugins = [...this.plugins, plugin];
    } else {
      this.plugins = [...this.plugins.slice(0, index), plugin, ...this.plugins.slice(index)];
    }
```

src/plugins/createPluginManager.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  public setup(): PluginSlots {
+    const pluginSlots: PluginSlots = {
+      executionDetailsTabs: [],
+    };
+    this.plugins.forEach(plugin => {
+      plugin.setup(pluginSlots);
```
rangoo94, line 19: Instead of getting raw objects, `plugin.setup` should get some utility to help to set it up, or this object would need to have helper methods.

As an example:
```ts
class PluginsConfiguration {
  public slots: PluginSlots = {
    executionDetailsTabs: [],
  };

  addTo(slotName: keyof PluginSlots, order?: number): void;
  addTo(slotName: string, order?: number): void;
  addTo(slotName: string, order?: number): void {
    // add an item to slot
  },
}

// [...]

public setup(): PluginsConfiguration {
  const config = new PluginsConfiguration();
  this.plugins.forEach(plugin => {
    plugin.setup(config);
  });
  return config;
}
```

src/components/molecules/ExecutionDetails/TestExecutionDetails/TestExecutionDetailsTabs.tsx
```diff
@@ -24,6 +27,7 @@ const LogOutput = lazy(() => import('@molecules').then(module => ({default: modu
 
 const TestExecutionDetailsTabs: React.FC = () => {
   const {data} = useContext(ExecutionDetailsContext);
+  const {pluginSlots} = useContext(PluginsContext);
```
rangoo94, line 30: It would be safer to have hooks for that, like:

```ts
const usePluginSlot = <T>(name: string, defaults?: T): T => {
  const {pluginSlots} = useContext(PluginsContext);
  return pluginSlots || defaults;
};
```

src/plugins/createPluginManager.ts
```diff
@@ -0,0 +1,28 @@
...
+  public constructor() {
+    this.plugins = [];
+  }
+
+  public add(plugin: Plugin): void {
+    plugin.order ? this.plugins.splice(plugin.order, 0, plugin) : this.plugins.push(plugin);
+  }
+
+  public setup(): PluginSlots {
+    const pluginSlots: PluginSlots = {
```
rangoo94, line 15: I wouldn't call these "slots" slots, these are rather some "items".

Besides, there are multiple cases to handle:
* Slots with a single item
* Slots with an array of items
* Data item
* Data lists
* Maybe: data object

src/plugins/utils.ts
```diff
@@ -0,0 +1,8 @@
+export const insertPluginsToArray = <T>(pluginsArray: any[], array: T[]): T[] => {
+  const components = [...array];
+  pluginsArray.forEach(plugin => {
+    const {priority, ...pluginData} = plugin;
+    priority ? components.splice(priority, 0, pluginData) : components.push(pluginData);
```
rangoo94, line 5: * Worth to consider unifying `priority`/`order` naming
* `if/else`,
* the comment from above for plugins `order`

src/plugins/utils.ts
```diff
@@ -0,0 +1,8 @@
+export const insertPluginsToArray = <T>(pluginsArray: any[], array: T[]): T[] => {
```
rangoo94, line 1: `pluginsArray` here should be either `unknown[]` or `T[]` (depending how strict we'd like to be)

src/plugins/createPluginManager.ts
haneabogdan: refactored the data model exactly like that, to have a render function, that can receive props, which can be passed to the components. This will be a very common use case, as we will need to pass contextual data each time (e.g. selected test execution, etc).

src/plugins/createPluginManager.ts
haneabogdan: for now, I just made a distinction between single items and array of items. Only difference for array, is that we need the helper function `insertPluginsToArray`, in order to insert plugin items between default components

src/App.tsx
```diff
@@ -108,6 +114,11 @@ const App: React.FC = () => {
   return composeProviders()
     .append(Suspense, {fallback: <Loading />})
     .append(StoreProvider, {})
+    .append(PluginsContext.Provider, {
+      value: {
+        pluginItems: pluginManager.setup(),
```
rangoo94, line 119: Probably should be memoized - otherwise each time there may be a new object created, re-rendering each view

src/components/molecules/ExecutionDetails/TestExecutionDetails/TestExecutionDetailsTabs.tsx
```diff
@@ -83,7 +87,7 @@ const TestExecutionDetailsTabs: React.FC = () => {
     }
   }, [isRunning, id]);
 
-  const items = [
+  const defaultExecutionDetailsTabs = [
     {
       key: 'LogOutputPane',
       label: 'Log Output',
```
rangoo94, line 93: Probably `Log Output` should receive high priority, maybe even `order: Infinity`

src/plugins/createPluginManager.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export class PluginManager {
+  private plugins: Plugin[];
+
+  public constructor() {
+    this.plugins = [];
+  }
+
+  public add(plugin: Plugin): void {
+    const order = plugin.order ?? -Infinity;
```
rangoo94, line 11: Optional: Alternatively, such logic could be extracted to some `OrderedList`, so the same logic could be applied in `insertPluginsToArray`

src/plugins/PluginsContext.ts
```diff
@@ -0,0 +1,24 @@
+import {createContext} from 'react';
+
+export interface PluginData {
+  getComponent: (props?: any) => any;
+  order?: number;
+}
+
+export type PluginItems = Record<string, PluginData[]>;
```
rangoo94, line 8: I don't think that these are `PluginItems`, but I don't have any good name for that; I'm not sure though if the name is even needed

src/plugins/PluginsContext.ts
```diff
@@ -0,0 +1,24 @@
+import {createContext} from 'react';
+
+export interface PluginData {
+  getComponent: (props?: any) => any;
```
rangoo94, line 4: Why the `PluginData` would have `getComponent`? `PluginData` should rather never be a component (more likely React element), and sometimes should be arrays or array items (i.e. tabs definition).

src/components/molecules/ExecutionDetails/TestExecutionDetails/TestExecutionDetailsTabs.tsx
```diff
@@ -24,6 +27,7 @@ const LogOutput = lazy(() => import('@molecules').then(module => ({default: modu
 
 const TestExecutionDetailsTabs: React.FC = () => {
   const {data} = useContext(ExecutionDetailsContext);
+  const {pluginItems} = useContext(PluginsContext);
```
rangoo94, line 30: We should rather have hooks for that, i.e.:

```ts
const usePluginDataList = <T[]>(name: string, defaults: T[] = []): T[] => {
  const {pluginItems} = useContext(PluginsContext);
  return pluginItems[name] || [];
};
```

This way it's safer, and i.e. such `usePluginDataList` hook could automatically handle the `insertToArray` and memoization

src/plugins/createPluginManager.ts
rangoo94: To pass contextual data, the components should use contexts instead; it's good to add support for providers too then

src/plugins/createPluginManager.ts
rangoo94: Plugin mechanism is basically an architecture foundation; it's better to have everything analyzed, covered and tested immediately - otherwise, we would need to anyway immediately do it. If foundation has some flaws and is used already throughout the application, it ends up with complex refactor - many moving parts that have to be redesigned again too

src/components/molecules/ExecutionDetails/TestExecutionDetails/TestExecutionDetailsTabs.tsx
haneabogdan: `order` prop is not supported within the `Tab` interface, hence we'll need to extend the typing for this (and probably other usecases in the future). Isn't it enough if we can insert tabs from plugin in between the default once, without having to set order for the defaults? Only downside I see is, that we can't enforce a certain tab to ALWAYS be the first

rangoo94: We cannot, as `order` will be unusable then. When you'll put `order: 1` to any element, it will go before every single one there. There is no way to order items *after* the defaults.

src/plugins/PluginsContext.ts
haneabogdan: We need a way to pass props to the injected components from the parent component (e.g. test execution id, status, etc). As we don't want to propagate them to the top of the component tree, I've added this render function that can be called within the parent component with the needed props and it will return the component with the needed state

src/components/molecules/ExecutionDetails/TestExecutionDetails/TestExecutionDetailsTabs.tsx
rangoo94: Besides, `order` doesn't have to be a part of prop - it may be a metadata of the item spec, like `{ order: 3, item: { key: '', label: '' }}`

src/plugins/PluginsContext.ts
rangoo94: It could be easily handled with Context providers/consumers.

rangoo94: By decentralized, I mean that in *core* you should give flexibility for the plugins. The *core* shouldn't care if the plugin will display some data as a modal, a tab, or in different window - *core* only has to provide essential data for the display.

Keeping it like `props` would mean that the plugin will have this data only in a single place, so it won't be simple to handle it throughout the plugin as a whole.

src/App.tsx
```diff
@@ -29,6 +29,8 @@ import {safeRefetch} from '@utils/fetchUtils';
 import {PollingIntervals} from '@utils/numbers';
 
 import {MessagePanelWrapper} from './App.styled';
+import {PluginManager} from './plugins/PluginManager';
```
rangoo94, line 32: You may use `@plugins` namespace

src/components/molecules/AiInsightsTab/AiInsightsTab.tsx
```diff
@@ -0,0 +1,8 @@
+import {usePluginState} from '@plugins/pluginHooks';
+
+const AiInsightsTab = () => {
+  const [aiInsightsProps] = usePluginState<{id: string}>('executionDetailsTabs');
```
rangoo94, line 4: Does the `AiInsightsTab` needs to know the current `id` of execution details tabs? It is enough to know that it's mounted (and it's possible to handle without state)

haneabogdan: it will need to know the `executionId`, because we will use it in the API request, to fetch AI insights specifically for this execution

rangoo94: Ahh, right, `id` mislead me 🙂 Probably worth to call it `executionId` then (and we could expose more data, including entity details)

src/components/molecules/AiInsightsTab/AiInsightsTab.tsx
```diff
@@ -0,0 +1,8 @@
...
+  return <div>This is a cloud only feature. Please log in to cloud in order to use this feature</div>;
```
rangoo94, line 5: Maybe:
```suggestion
  return <div>This is a Cloud-only feature. Please [link]connect your cluster to Testkube Cloud[/link] in order to use this feature.</div>;
```

src/plugins/interfaces.ts
```diff
@@ -0,0 +1,11 @@
+import PluginScope from './PluginScope';
```
rangoo94, line 1: `types.ts` could be a better name than `interfaces.ts`, as you may want to have mapped types here too

src/plugins/interfaces.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export interface Plugin {
+  name: string;
+  setup: (scope: PluginScope) => void;
+  order?: number;
+}
+
+export interface PluginMetaData {
```
rangoo94, line 9:
```suggestion
export interface SlotMetadata {
```

src/plugins/pluginHooks.ts
```diff
@@ -0,0 +1,17 @@
+import {useCallback, useContext, useMemo} from 'react';
+
+import PluginsContext from './PluginsContext';
+import {orderArray} from './utils';
+
+export const usePluginState = <T>(name: string) => {
+  const {scope} = useContext(PluginsContext);
+  const setStateCallback = useCallback((value: any) => scope.setState.bind(name, value), [name, scope]);
```
rangoo94, line 8: It should be either:
```suggestion
  const setStateCallback = useMemo(() => scope.setState.bind(null, name), [name, scope]);
```
Or:
```suggestion
  const setStateCallback = useCallback(scope.setState.bind(null, name), [name, scope]);
```

src/plugins/utils.ts
```diff
@@ -0,0 +1,15 @@
+export const orderArray = (array: any[]) => {
+  const result: any[] = [];
+  array.forEach((item: any) => {
+    if (!item) {
+      return;
+    }
+    const index = result.findIndex(x => x.metadata.order < item.metadata.order);
```
rangoo94, line 7: Metadata are optional:
```suggestion
    const index = result.findIndex(x => (x.metadata?.order ?? -Infinity) < (item.metadata?.order ?? -Infinity));
```

src/components/molecules/AiInsightsTab/AiInsightsTab.tsx
haneabogdan: test will change with the next PR which contains the actual implementation of the AI Insights. for now, this is a placeholder

src/App.tsx
```diff
@@ -37,7 +40,11 @@ const Sources = lazy(() => import('@pages').then(module => ({default: module.Sou
 const Triggers = lazy(() => import('@pages').then(module => ({default: module.Triggers})));
 const GlobalSettings = lazy(() => import('@pages').then(module => ({default: module.GlobalSettings})));
 
-const App: React.FC = () => {
+export interface AppProps {
+  pluginManager: PluginManager;
```
rangoo94, line 44: Actually, I'm thinking about the community plugins and I've got a struggle. There are two options we could approach it:

* Plugins may be injected only for Dashboard context
* Plugins may be injected for both Dashboard and Cloud context

if we want it only for a dashboard, we could pass `plugins: Plugin[]` here instead of `pluginManager`. This would make that implementation internal, simplifying the dependencies. `App` should then instantiate and setup `PluginManager`.

If we want to allow injecting also in the Cloud context in the future, we should pass `plugins: PluginScope` (btw, I'm not sure if this name is appropriate for that, but it's fine). Otherwise, we would call `manager.setup()` twice - in Cloud context (before Dashboard is loaded) and in Dashboard context, while it should be setup only once.

The first solution is simpler, but the second one not much harder, yet may be helpful in the future. What are your thoughts on that?

src/components/atoms/NewTab/NewTab.styled.tsx
```diff
@@ -0,0 +1,6 @@
+import styled from 'styled-components';
+
+export const NewTabContainer = styled.div`
+  display: flex;
+  gap: 8px;
```
rangoo94, line 5: Optionally, wouldn't just margin left and middle vertical align on icon be simpler?

src/App.tsx
rangoo94: The second solution means that we may have both plugins for Dashboard (i.e. Test Execution) and Cloud (i.e. Environment, or something that just works all the time).

Because of that, you would need to call `manager.setup()` in the Cloud's `App` too - otherwise, it would be only instantiated in Dashboard.

haneabogdan: now I get what you're saying. Yes, that's true. If we also aim to have cloud plugins, we would have to call `manager.setup()` twice. It's an interesting use case to have also cloud plugins, but personally I think it's way down the road. Also, it adds another layer of complexity, since these users don't have access to the cloud repo and most likely also the plugins shouldn't be publicly available. We would need to think of an import mechanism.  Only downside I see if we decide not to do it now and only much later, is that we would need to refactor all the other places where plugins are being used. **I would still stick to OSS plugins for the time being.**

rangoo94: > Also, it adds another layer of complexity, since these users don't have access to the cloud repo and most likely also the plugins shouldn't be publicly available.

The Users won't need access to the Cloud repo, they only would need to know the interface.

> We would need to think of an import mechanism.

For community plugins, even for OSS Dashboard, we still need to think about the import mechanism.

> Only downside I see if we decide not to do it now and only much later, is that we would need to refactor all the other places where plugins are being used. I would still stick to OSS plugins for the time being.

That was my only doubt. We could go with the 1st solution with `plugins: Plugin[]` then 👍

src/AppRoot.tsx
```diff
@@ -103,6 +112,25 @@ const AppRoot: React.FC = () => {
     [navigate, location]
   );
 
+  const plugins: Plugin[] = [
```
rangoo94, line 115: We could memoize it without dependencies (`useMemo(() => [{name: 'ai-insights'/*...*/}], [])`)

src/components/molecules/ExecutionDetails/TestExecutionDetails/TestExecutionDetailsTabs.tsx
```diff
@@ -43,14 +47,18 @@ const TestExecutionDetailsTabs: React.FC = () => {
...
 
   const isRunning = useIsRunning(status);
 
   const decomposedVars = decomposeVariables(variables || {});
 
   const whetherToShowArtifactsTab = executorsFeaturesMap[testType]?.includes('artifacts');
 
+  useEffect(() => {
+    setTestExecutionTabsData({execution, test: entityDetails});
+  }, [id]);
```
rangoo94, line 60: It should update on `entityDetails`/`execution` change too - otherwise we will not receive updates.

src/components/molecules/ExecutionDetails/TestExecutionDetails/TestExecutionDetailsTabs.tsx
```diff
@@ -83,39 +91,64 @@ const TestExecutionDetailsTabs: React.FC = () => {
...
+      metadata: {
+        order: 2,
+      },
     },
     decomposedVars.length
       ? {
-          key: 'Variables',
-          label: 'Variables',
-          children: <ExecutionsVariablesList variables={decomposedVars} />,
+          value: {
```
rangoo94, line 139: This whole array should be filtered out of nullish values (`.filter(Boolean)`), otherwise `orderArray` will fail

rangoo94 (approved):

rangoo94 (approved):

PR #796, opened 2023-07-25
tech: add help utilities for the Zustand store

rangoo94:
## Changes

* Add hooks for the Zustand store that allows to easily manage the data ~and their publicity~
* Make the Triggers slice/store standalone
* Use these new hooks for the Triggers store

## Brief

I added a few helpful utilities for our Zustand stores. It will make the work with Zustand much easier and type-safe. The implementation is pretty complex, so I have added documentation for that [**here**](https://github.com/kubeshop/testkube-dashboard/blob/dawid/tech/add-visibility-store-utils/src/store/README.md).

I will shortly describe the available features.

### `useStore`

It's working exactly as before - allows computing some data from the state.

### `useStorePick`

It simplifies taking selected attributes from the store.

```ts
// Now:
const {userData, userToken} = useStorePick('userData', 'userToken');

// Before:
const {userData, userToken} = useStore(state => ({
  userData: state.userData,
  userToken: state.userToken,
}));
```

### `useStoreSync`

Similarly to the `useStoreSetter`, this hook will call corresponding setters of the object values, when the value is different.

```ts
// Slice example

interface FooSlice {
  foo: string;
  bar: string;
  updates: number;
  setBar: (bar: string) => void;
}

const createFooSlice: StateCreator<FooSlice> = set => ({
  foo: '',
  bar: '',
  updates: 0,
  setBar: (bar: string) => set(state => ({bar, updates: state.updates + 1})),
});

// Usage example

useStoreSync({
  foo: 'value', // If `foo` will be different than "value", it will set `foo` to "value"
  bar: 'value', // If `bar` will be different than "value", it will call `setBar("value")`
});
```

### `useStoreField`

It is a common interface, that returns the React-like interface for reading and updating the value.

When it will find a setter in a format of `set<PROPERTY_NAME>`, it will call it. Otherwise, it will just replace the value in state "magically".

```ts
// Slice example

/* Look above */

// Usage example

const [foo, setFoo] = useStoreField('foo');
const [bar, setBar] = useStoreField('bar');
```

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test


pavloburchak:
Looking at the implementation and features added all seems impressive to me. But I doubt that this custom hooks are actually easier to use then previous approach and doubt it have any benefits, except of few lines of code removed when getting state. Maybe I am missing something? 

Also have a question to a maintainability of the code. It is pretty hard to understand what is going on in this types. What if we add them everywhere across the project and then some bug will appear or we need to modify them? Personally I am not sure that I will be able to fix that quickly, it is just seems safer to use native hooks, like `useStore`.

rangoo94:
@pavloburchak, I will start from answering the second part of the comment, and then will go back to beginning.

> Also have a question to a maintainability of the code. It is pretty hard to understand what is going on in this types. What if we add them everywhere across the project and then some bug will appear or we need to modify them?

I could agree with that. The implementation right now is quite complex. The code complexity is mostly because of `Internal<T>`, `NonPublic<T>`, and `NonPublicWrite<T>` types. On the one hand, it gives a nice option to encapsulate the code, but on the other, we could as well "enforce" it manually during the code review.

Additionally, to simplify the implementation, we could avoid having abstract hooks (i.e. `useStoreState(store, selector)`) in favor of hooks for specific stores (`useTriggers(selector)`).

> I doubt that this custom hooks are actually easier to use then previous approach and doubt it have any benefits, except of few lines of code removed when getting state. Maybe I am missing something?

When it comes to reading (`use<*>Pick`), there is not a lot of benefits:
* IDE will help to autocomplete the names,
* We avoid having 1-2 lines of code for each property

Nothing special, yet it's a pretty small chunk of code.

But when we take writing to the store under consideration, the situation is quite different. Without having `use<*>Setter`, `use<*>Sync` we will end up duplicating many times pretty complex logic, risking mistakes and problematic refactors. I'll show some simple diff examples.

To make the comparison simpler:
* I'll ignore the `createSliceFactory` because it won't be needed if we'll avoid `Internal<T>`-like types,
* I'll assume a standalone store for code clarity, but it could still work with the current solution (one store, multiple slices)

#### `use<*>Setter`

This hook simplifies managing the state a lot. Adding a new field to the slice is much simpler, although still far from `use<*>Sync` or `use<*>Field`. It could be treated as a low-level interface though.

```diff
diff --git @store/triggers

 import {StateCreator} from 'zustand';
 import {TestTrigger, TriggersKeyMap} from '@models/triggers';

 interface TriggersState {
   keyMap?: TriggersKeyMap;
   current?: string;
-  setKeyMap: (triggersKeyMap: TriggersKeyMap) => void;
-  setCurrent: (trigger: string) => void;
 }

 export const createTriggersSlice: StateCreator<TriggersSlice> = set => ({
   keyMap: undefined,
   current: undefined,
-  setKeyMap: (newKeyMap: TriggersKeyMap) => set({keyMap: newKeyMap}),
-  setCurrent: (newTrigger: TestTrigger) => set({currentr: newTrigger}),
 });

diff --git @components/example

 import {useEffect} from 'react';
 import {useParams} from 'react-router-dom';
-import {useTriggers} from '@store/triggers';
+import {useTriggers, useTriggersSetter} from '@store/triggers';

 const Example: React.FC = () => {
   const {id} = useParams();
-  const {current, setCurrent} = useTriggers(state => ({
-    current: state.current,
-    setCurrent: state.setCurrent,
-  });
+  const current = useTriggers(state => state.current);
+  const setCurrent = useTriggersSetter('current');

   useEffect(() => {
     setCurrent(id);
   }, [id]);
 };
```

#### `use<*>Field`

`use<*>Field` is a small helper, that exposes data from `use<*>` and `use<*>Setter` in a convenient manner. This one is the least important, yet the simplest one, and handles a pretty common use-case - `value`/`onChange` pair.

```diff
diff --git @store/triggers

 import {StateCreator} from 'zustand';
 import {TestTrigger, TriggersKeyMap} from '@models/triggers';

 interface TriggersState {
   keyMap?: TriggersKeyMap;
   current?: string;
-  setKeyMap: (triggersKeyMap: TriggersKeyMap) => void;
-  setCurrent: (trigger: string) => void;
 }

 export const createTriggersSlice: StateCreator<TriggersSlice> = set => ({
   keyMap: undefined,
   current: undefined,
-  setKeyMap: (newKeyMap: TriggersKeyMap) => set({keyMap: newKeyMap}),
-  setCurrent: (newTrigger: TestTrigger) => set({currentr: newTrigger}),
 });

diff --git @components/example

 import {useEffect} from 'react';
 import {useParams} from 'react-router-dom';
-import {useTriggers} from '@store/triggers';
+import {useTriggersField} from '@store/triggers';

 const Example: React.FC = () => {
   const {id} = useParams();
-  const {current, setCurrent} = useTriggers(state => ({
-    current: state.current,
-    setCurrent: state.setCurrent,
-  });
+  const [current, setCurrent] = useTriggersField('current');

   useEffect(() => {
     setCurrent(id);
   }, [id]);
 };
```

#### `use<*>Sync`

When it comes to `use<*>Sync`, it handles much more problematic case - when you want to sync local or external data with the store.

```diff
diff --git @store/triggers

 import {StateCreator} from 'zustand';
 import {TestTrigger, TriggersKeyMap} from '@models/triggers';

 interface TriggersState {
   keyMap?: TriggersKeyMap;
   current?: string;
-  setKeyMap: (triggersKeyMap: TriggersKeyMap) => void;
-  setCurrent: (trigger: string) => void;
 }

 export const createTriggersSlice: StateCreator<TriggersSlice> = set => ({
   keyMap: undefined,
   current: undefined,
-  setKeyMap: (newKeyMap: TriggersKeyMap) => set({keyMap: newKeyMap}),
-  setCurrent: (newTrigger: TestTrigger) => set({currentr: newTrigger}),
 });

diff --git @components/example

 import {useEffect} from 'react';
 import {useParams} from 'react-router-dom';
import {useGetKeyMap} from '@services/triggers';
-import {useTriggers} from '@store/triggers';
+import {useTriggersSync} from '@store/triggers';

 const Example: React.FC = () => {
   const {id} = useParams();
   const {data: keyMap} = useGetKeyMap(id);
-  const {current, setCurrent, keyMap, setKeyMap} = useTriggers(state => ({
-    current: state.current,
-    setCurrent: state.setCurrent,
-    keyMap: state.keyMap,
-    setKeyMap: state.setKeyMap,
-  });
-
-  useEffect(() => {
-    setCurrent(id);
-  }, [id]);
-
-  useEffect(() => {
-    setKeyMap(keyMap);
-  }, [keyMap]);
+  useTriggersSync({current: id, keyMap});
 };
```

What do you think about that?

Cc: @pavloburchak, @mortada-codes, @haneabogdan, @Toxer10

pavloburchak:
Understood. I guess when we will continue changing stores to Zustand, more understanding will come. I can see the improvement now, we are getting rid of a lot of boilerplate code during store creation, so this is good stuff.

Still, let's see how it goes with our next store and possible improvement or disadvantage will be visible by itself. Approved!

pavloburchak (approved):

rangoo94:
@pavloburchak, @mortada-codes, @EduardIvanov22, @haneabogdan - I have updated the PR to delete global store access hooks, and the visibility types (`Internal<T>` etc). It has reduced the complexity significantly.

Please review if you can 🙂

mortada-codes (approved):

haneabogdan (approved):

rangoo94:
Hidden the `useStoreSetter` hook, as it won't be commonly used, and `useStoreField` could be used to get the setter anyway:

```ts
const [,setCurrent] = useTriggersField('current');
```

PR #797, opened 2023-07-26
feat: E2E tests - test creation - File and String

tkonieczny:


.eslintignore
```diff
@@ -11,3 +11,4 @@ server.js
 build.js
 init.js
 site
+test/dashboard-e2e/fixtures/files
```
rangoo94, line 14: trivial: missing new line on EOF

.prettierignore
```diff
@@ -5,3 +5,4 @@ build
 node_modules
 playwright-report
 test-results
+dashboard-e2e/fixtures/files
```
rangoo94, line 8: trivial: missing new line on EOF

test/dashboard-e2e/config.ts
```diff
@@ -1,5 +1,9 @@
+import dotenv from 'dotenv';
```
rangoo94, line 1: Is the `dotenv` required here? I've seen that you deleted it in the last Cloud PR.

test/dashboard-e2e/helpers/common.ts
```diff
@@ -12,13 +13,25 @@ export function validateTest(testData: Partial<TestData>, createdTestData: TestD
...
+
     Object.keys(testData.content.repository).forEach(key => {
       expect(testData.content.repository[key]).toEqual(createdTestData.content.repository[key]);
     });
+  } else if (testData.content.type === 'string') {
+    expect(testData.content.type).toEqual(createdTestData.content.type);
+
+    Object.keys(testData.content).forEach(key => {
+      expect(testData.content[key]).toEqual(createdTestData.content[key]);
+    });
```
rangoo94, line 29:
```suggestion
     expect(createdTestData.content).toMatchObject(testData.content);
```

test/dashboard-e2e/helpers/common.ts
```diff
@@ -12,13 +13,25 @@ export function validateTest(testData: Partial<TestData>, createdTestData: TestD
...
   // testSources
   if (testData.content.type === 'git') {
+    expect(testData.content.type).toEqual(createdTestData.content.type);
+
     Object.keys(testData.content.repository).forEach(key => {
       expect(testData.content.repository[key]).toEqual(createdTestData.content.repository[key]);
     });
+  } else if (testData.content.type === 'string') {
+    expect(testData.content.type).toEqual(createdTestData.content.type);
+
```
rangoo94, line 26: This is checked below anyway.
```suggestion
```

test/dashboard-e2e/package.json
```diff
@@ -11,7 +11,7 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "@playwright/test": "1.32.3",
+    "@playwright/test": "^1.32.3",
```
rangoo94, line 14: Shouldn't we stick to `1.32.3`?

test/dashboard-e2e/pages/CreateTestPage.ts
```diff
@@ -35,13 +39,21 @@ export class CreateTestPage {
...
         }
       }
-    } else {
-      throw Error('Type not supported by selectTestSource - extend CreateTestPage');
+    } else if (type === 'string') {
+      await this.setSelectionSearch('String', 'testSource');
+      await this.setBasicInput(contentData.data, 'string');
+    } else if (type === 'file') {
+      await this.setSelectionSearch('File', 'testSource');
+      // await this.setBasicInput(contentData.data, 'string');
```
rangoo94, line 47:
```suggestion
```

test/dashboard-e2e/pages/CreateTestPage.ts
```diff
@@ -65,4 +77,8 @@ export class CreateTestPage {
   private async clickCreateTestButton(): Promise<void> {
     await this.page.click('button[data-test="add-a-new-test-create-button"]');
   }
+
+  private static getAbsoluteFixtureFilePath(fixtureFileName: string) {
+    return path.resolve(`fixtures/files/${fixtureFileName}`);
```
rangoo94, line 82: Optionally: This would be relative from working directory while executing tests, selecting based on the current path could be more stable:
```suggestion
    return path.resolve(__dirname, `../fixtures/files/${fixtureFileName}`);
```

rangoo94 (commented):
Something is strange with executing the E2E tests for this PR - Playwright tests have apparently finished, but the test execution is still not finished.

test/dashboard-e2e/config.ts
tkonieczny: I deleted it from playwright config (moved it to config.ts where ENVs are currently loaded)

test/dashboard-e2e/helpers/common.ts
tkonieczny: I iterate over `testData.content` keys on purpose here - response still contain redundant things (for example, empty repository) even if a test is created with string type

test/dashboard-e2e/config.ts
tkonieczny: I moved it from `playwright.config.ts` to `config.ts` because otherwise ENVs from `.env` file weren't loaded correctly

test/dashboard-e2e/helpers/common.ts
rangoo94: `expect(obj1).toMatchObject(obj2)` is checking if all of the properties in `obj2` are matching in the `obj1`, so all not recognized properties will be ignored.

test/dashboard-e2e/config.ts
rangoo94: Ahh, that's right - `playwright.config.ts` is loading the `config.ts` before adjusting the envs 👍

tkonieczny:
> E2E tests failed with server crashed: `× The run has failed: testkube api server crashed`

Looks like it was caused with playwright serving html reports again - it caused the pod to be stuck running forever.
```  Serving HTML report at http://localhost:9323. Press Ctrl+C to quit.
Failed to open browser on http://localhost:9323```

I'm not sure what exactly changed, but looks like CI=1 isn't set by default anymore.

**Comment by rangoo94** [comment]
> Looks like it was caused with playwright serving html reports again - it caused the pod to be stuck running forever.

Would it crash the whole API Server pod?

> I'm not sure what exactly changed, but looks like CI=1 isn't set by default anymore.

I think that I deleted it from the Executor because we don't add `CI=1` to other executors too

**Comment by rangoo94** [comment]
> CI env (and CI mode in general) isn't something Playwright-specific - it's widely used across different tools and providers. So, we should probably also set it by default for all official executors.

We don't use it in other executors, therefore we should rather either consistently use it everywhere, or don't use it anywhere.

I would suggest adding `CI=1` automatically on the job level (for all jobs), not the executor or test level.

**Comment by tkonieczny** [comment]
Added an issue for that: https://github.com/kubeshop/testkube/issues/4214

**Review by rangoo94 — APPROVED** [review]

**Comment by rangoo94** [comment]
@tkonieczny - looks like it's [**failing**](https://dev.testkube.io/tests/executions/dashboard-e2e-tests/execution/64c7afb81f07b57164b2aa62) for file-based K6 test:
```
  1 failed
    [chromium] › create-test.spec.ts:16:7 › Creating test k6-file ──────────────────────────────────
  30 skipped
  12 passed (2.2m)
```

tkonieczny:
@rangoo94 took me a while to figure out what happens here,  but I was able to reproduce it.

It's possible to create a test using vercel preview `git`, or `string`. But, when I tried to do it for `file` it failed:
```Access to fetch at 'https://dev.testkube.io/results/v1/tests' from origin 'https://testkube-dashboard-o4mm4626z-testkube-cloud.vercel.app' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.```
Screenshot from 2023-07-31 23-31-12

tkonieczny:
CORS now works fine again. The second issue was Vercel preview's "live feedback" iframe that was displayed on top of the "test create" button (which wasn't visible in traces).

PR #807, opened 2023-08-02
feat: add test suite title plugin slot

haneabogdan:
This PR...

- creates new plugin slot for TestSuite title icon
- creates new plugin hook to fetch single component for slot
- creates new Tag component

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/models/entity.ts
+++ b/src/models/entity.ts
@@ -12,7 +12,7 @@ export type Entity = 'test-suites' | 'tests';
 export type EntityListBlueprint = {
   pageTitle: string;
 
-  pageTitleIcon?: ReactNode;
+  pageTitleAddon?: ReactNode;
 
   pageDescription?: any;
 
--- a/src/models/entity.ts
+++ b/src/models/entity.ts
@@ -1,4 +1,4 @@
-import {FC} from 'react';
+import {FC, ReactNode} from 'react';
 
 import {ModalConfig} from '@contexts/ModalContext';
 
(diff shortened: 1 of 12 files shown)


src/plugins/pluginHooks.ts
```diff
@@ -10,6 +10,11 @@ export const usePluginState = <T>(name: string) => {
   return pair;
 };
 
+export const useSinglePluginSlot = (name: string) => {
```
rangoo94, line 13: * Maybe `usePluginSlot` or `usePluginSingleSlot`, to keep it prefixed with `usePlugin` namespace.
   * `usePluginSlot` makes rather more sense, as `usePluginSlotList` is specific, claiming that "single" is default
* It should either ensure that there is only single element, or use priority - `const usePluginSlot = (name: string) => usePluginSlotList(name)[0];`

src/models/entity.ts
```diff
@@ -12,6 +12,8 @@ export type Entity = 'test-suites' | 'tests';
 export type EntityListBlueprint = {
   pageTitle: string;
 
+  pageTitleIcon?: ReactNode;
```
rangoo94, line 15: Maybe `pageTitleAddon`, given that it's not necessarily the icon?

rangoo94 (approved):

PR #811, opened 2023-08-03
fix: plugin visibility

haneabogdan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


Toxer10 (approved):

src/assets/images/lockedIcon.svg
rangoo94, line 1: Is that icon expected within this PR?

src/plugins/pluginHooks.ts
```diff
@@ -18,7 +18,12 @@ export const usePluginSlotList = (name: string, defaults: any[] = []) => {
   // by default, all elements are visible,
   // unless visibility is configured explicitly in the metadata
   const elements = useMemo(
-    () => orderArray([...defaults, ...scope.getSlot(name)].filter(x => !x.metadata?.visible || x.metadata.visible())),
+    () =>
+      orderArray(
+        [...defaults, ...scope.getSlot(name)].filter(
+          x => !x.metadata?.visible || x.metadata.visible(scope.getState(name))
```
rangoo94, line 24: Wasn't `metadata.visible` without params?

haneabogdan: needed state within visibility. but we can get it directly in the plugin definition, since we have access to `scope.getState`. no need for param

src/assets/images/lockedIcon.svg
haneabogdan: moved to OSS since we will need it for this feature. not the best place to mix these files together. will push it with the OSS version of AI insights

haneabogdan:
closing PR. not needed

src/plugins/pluginHooks.ts
rangoo94: Exactly :) I double-checked, and `visible` doesn't contain `params`, so this change is not necessary.

PR #813, opened 2023-08-03
feat: AI insights tab

haneabogdan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mortada-codes (approved):

src/plugins/definitions/ai-insights.tsx
```diff
@@ -0,0 +1,30 @@
+import {IconLabel, Tag} from '@atoms';
+
+import {AiInsightsTab} from '@molecules';
+
+import {Plugin, TestExecutionTabsInterface} from '@plugins/types';
+
+const getAiInsightsPlugin = (): Plugin => ({
```
rangoo94, line 7: Maybe `createAiInsightsPlugin` would be better, and same for its imports

src/components/molecules/AiInsightsTab/AiInsightsTab.tsx
```diff
@@ -1,19 +1,40 @@
...
-      This is a cloud only feature. Please log in to cloud in order to use this feature for id
-      {execution.id}
-    </div>
+    <AiInsightContainer>
+      <AiInsightContent>
+        <LockedIcon />
+        <Text className="big bold"> This feature is available only in Testkube Cloud.</Text>
+        <Text color={Colors.slate400}>
+          Start using Testkube Cloud to get AI insights for your test executions, as well as other exclusive features.
+          Learn more
```
rangoo94, line 26: Should it be there? Or maybe there is some text missing?

```suggestion
```

src/utils/externalLinks.ts
```diff
@@ -10,6 +10,7 @@ export enum externalLinks {
   sourcesApi = 'https://docs.testkube.io/openapi/#tag/test-sources',
   sourcesDocumentation = 'https://docs.testkube.io/articles/creating-tests/#test-source',
   testSources = 'https://docs.testkube.io/articles/test-sources',
+  testkubeCloud = 'https://cloud.testkube.io',
```
rangoo94, line 13: Could be next to `OSStoCloudMigration`, to keep Cloud links near

rangoo94 (commented):
Something is wrong with padding, but it looks good when the screen height is smaller

Zrzut ekranu 2023-08-3 o 15 06 34

src/components/molecules/AiInsightsTab/AiInsightsTab.tsx
haneabogdan: should be a link to an explanatory page. Since we don't have the page now, this is a placeholder

haneabogdan:
@rangoo94 can you detail what is the problem with the padding?

src/components/molecules/AiInsightsTab/AiInsightsTab.tsx
haneabogdan: good point. will change it to an empty link. so that we don't forget about it

rangoo94:
> @rangoo94 can you detail what is the problem with the padding?

The padding on the top and bottom are different. That's because `min-height: 25vh` - I think that it should be only in case there is an actual log displayed, not for consent / redirect message.

src/components/molecules/AiInsightsTab/AiInsightsTab.tsx
rangoo94: For now it could just redirect to Cloud 🙂

rangoo94:
>  weirdly enough, I can't see any differences with/without it. will merge the PR for now and analyse this further

Maybe you have small window or DevTools opened?

PR #815, opened 2023-08-03
feat: add routing to different tabs

rangoo94:
## Changes

- add routing to specific tabs in the UI
- clean up routes for tests and test suites, while maintaining backwards compatibility

## Fixes

- kubeshop/testkube#3708

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/DashboardRewrite.tsx
+++ b/src/DashboardRewrite.tsx
@@ -1,0 +1,23 @@
+import {FC, useContext, useLayoutEffect} from 'react';
+import {useParams} from 'react-router-dom';
+
+import {DashboardContext} from '@contexts';
+
+interface DashboardRewriteProps {
+  pattern: string;
+  keepQuery?: boolean;
+}
+
+const DashboardRewrite: FC<DashboardRewriteProps> = ({pattern, keepQuery = false}) => {
+  const params = useParams();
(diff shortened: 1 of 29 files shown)


pavloburchak:
Looks good, but I spotted one small issue. Maybe we should set route for default tabs as well? Or for example redirect from `/tests/some-test/` to `/tests/eduard-test/executions`. Seems not really great, when you have multiple.

https://github.com/kubeshop/testkube-dashboard/assets/25755402/1a649593-50d9-4386-9995-57e1ad20a317

pavloburchak (approved):

rangoo94:
> Looks good, but I spotted one small issue. Maybe we should set route for default tabs as well? Or for example redirect from /tests/some-test/ to /tests/eduard-test/executions. Seems not really great, when you have multiple.

We could do so, but I'm not sure how to handle then the case of generic settings (like it's implemented now for executors, test sources and triggers, with `:settingsTab`). I think that it's a valid point, but I would get back to it after all the pages will be composed properly, as it may happen that the mechanism will look differently then. 👍

PR #816, opened 2023-08-04
feat: Your testkube instalation block

pavloburchak:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/store/apiDetails.ts
rangoo94, line 1: I think that `clusterDetails` could be more precise

src/store/apiDetails.ts
```diff
@@ -0,0 +1,20 @@
+import {StateCreator} from 'zustand';
+
+import {connectStore, createStoreFactory} from '@store/utils';
+
+export interface ApiDetailsSlice {
+  data: any;
```
rangoo94, line 6: Instead of having `data`, we can have all the details directly there

src/components/pages/GlobalSettings/General/YourInstallation/YourInstallationItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+import Colors from '@styles/Colors';
+
+import {YourInstallationItemContainer} from './YourInstallation.styled';
+
+type YourInstallationItemProps = {
+  label: string;
+  value?: string;
+};
+
+const YourInstallationItem: React.FC<YourInstallationItemProps> = props => {
```
rangoo94, line 12: You may use `memo`

src/components/pages/GlobalSettings/General/YourInstallation/YourInstallation.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import env from '../../../../../env';
+
+import {YourInstallationContainer} from './YourInstallation.styled';
+import YourInstallationItem from './YourInstallationItem';
+
+const YourInstallation: React.FC = () => {
+  const [data] = useApiDetailsField('data');
+
+  const versionRegex = /^(\d+\.)?(\d+\.)?(\*|\d+)$/;
```
rangoo94, line 15: * May be worth to allow`v` on the beginning (`v?`)
* We support tagged (like alpha, beta) and dirty versions (basically - instead of `$`, `([.-].+)?$` would be fine
* Why do we allow `*`, `1.*` and `1.*.*` as a version?

src/components/pages/GlobalSettings/General/YourInstallation/YourInstallation.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import {useApiDetailsField} from '@src/store/apiDetails';
+
+import env from '../../../../../env';
+
+import {YourInstallationContainer} from './YourInstallation.styled';
+import YourInstallationItem from './YourInstallationItem';
+
+const YourInstallation: React.FC = () => {
+  const [data] = useApiDetailsField('data');
```
rangoo94, line 13: You will be not setting here, so you may use `useApiDetailsPick`:
```suggestion
  const {data} = useApiDetailsPick('data');
```

As it will allow multiple values, like:
```suggestion
  const {data, version} = useApiDetailsPick('data', 'version');
```

src/components/pages/GlobalSettings/General/YourInstallation/YourInstallation.tsx
pavloburchak: 1. Api sending without it, I don't think so
2. this is not working. Modified so it supports dirty versions
3. Updated

rangoo94: Regarding (2) - what is not working? Besides, should we actually check the version at all, instead of showing it always when we receive it?

pavloburchak: In case of api version, sometimes it is sending a commit, in that case we want to display # before it. We might not check for other fields

rangoo94: Why we wouldn't want to show the commit? We tag the images with commit i.e. for DEV version. Besides, you may as well use SHA hash for specific Docker images too.

pavloburchak: I don't understand the problem here. I just display the version, in case it is not formatred like version, I display it with #, because it is a commit

src/App.tsx
```diff
@@ -101,16 +103,18 @@ const App: React.FC<AppProps> = ({plugins}) => {
...
-      // Handle race condition
-      if (getApiEndpoint() !== apiEndpoint) {
-        return;
-      }
-
-      // Display popup
-      notificationCall('failed', 'Could not receive data from the specified API endpoint');
-      setEndpointModalState(true);
-    });
+    getApiDetails(apiEndpoint)
```
rangoo94, line 106: This effect is called only in case the `REACT_APP_API_SERVER_ENDPOINT` is not set. When it will be set, it won't run:

https://testkube-dashboard-git-pavlo-display-version-testkube-cloud.vercel.app/settings?~api_server_endpoint=https://dev.testkube.io/results/v1

rangoo94: * The best solution would be to replace this `useEffect` with `useAsync` and `useEffect`,
* The easiest way would be to move `isApiEndpointLocked()` condition to the condition below:
  * `location.pathname === '/apiEndpoint'` instead of `location.pathname === '/apiEndpoint' || isApiEndpointLocked()`
  * `!apiEndpoint && !isApiEndpointLocked()` instead of `!apiEndpoint`

pavloburchak: done

src/App.tsx
```diff
@@ -92,25 +94,29 @@ const App: React.FC<AppProps> = ({plugins}) => {
...
+      .then(setClusterDetails)
+      .catch(() => {
+        // Handle race condition
+        if (getApiEndpoint() !== apiEndpoint) {
+          return;
+        }
+
+        // Display popup
+        notificationCall('failed', 'Could not receive data from the specified API endpoint');
+        if (isApiEndpointLocked()) {
```
rangoo94, line 116:
```suggestion
        if (!isApiEndpointLocked()) {
```

src/components/pages/GlobalSettings/General/YourInstallation/YourInstallation.tsx
```diff
@@ -0,0 +1,37 @@
+import {memo} from 'react';
+
+import {FullWidthSpace} from '@custom-antd';
+
+import {ConfigurationCard} from '@molecules';
+
+import {useClusterDetailsPick} from '@store/clusterDetails';
+
+import env from '../../../../../env';
```
rangoo94, line 9: Maybe:
```suggestion
import env from '@src/env';
```

src/store/clusterDetails.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import {ApiDetails} from '@services/apiEndpoint';
+
+import {connectStore, createStoreFactory} from '@store/utils';
+
+export interface ClusterDetailsSlice extends ApiDetails {
+  reset: () => void;
+  setClusterDetails: (state: ApiDetails) => void;
+}
+const initialState = {commit: '', context: '', helmchartVersion: '', namespace: '', version: '', url: ''};
```
rangoo94, line 11: Why is this extracted to separate variable? It's used only once, in the actual initial state creation, so probably better to just keep it in place.

src/services/apiEndpoint.ts
```diff
@@ -122,25 +128,29 @@ export async function getApiDetails(apiEndpoint: string): Promise<ApiDetails> {
...
 }
 
 export function useUpdateApiEndpoint(): (apiEndpoint: string) => Promise<boolean> {
   const {dispatch} = useContext(MainContext);
+  const {reset: resetClusterDetailsState, setClusterDetails} = useClusterDetailsPick('reset', 'setClusterDetails');
 
   return useMemo(
     () => async (apiEndpoint: string) => {
       const prevApiEndpoint = getApiEndpoint();
+      resetClusterDetailsState();
```
rangoo94, line 141: Instead of resetting it this way, you may pass `deps` to `initializeClusterDetailsStore`:
```ts
  const [ClusterDetailsProvider] = initializeClusterDetailsStore(undefined, [apiEndpoint]);
```

pavloburchak: This is not working as expected

src/store/clusterDetails.ts
pavloburchak: It used twice, initially and in reset

src/services/apiEndpoint.ts
rangoo94: What do you mean? What is the unexpected behavior? It should work out of the box.

pavloburchak: I assume that state is reseting after I apply new values. So I get empty fields in a result

rangoo94: Isn't it working the same way now?

pavloburchak: No, I firstly reset the state, and then add set values from new endpoint

rangoo94: I think that I understood what was happening. I'll take a look at it and adjust the logic.

pavloburchak: @rangoo94  So, can we merge this PR. Seems like a problem not really related to the feature

rangoo94: The problem is in this PR, not in the mechanism of initializing the store 🙂 The initialization works fine, we have incorrectly defined it there though. I will fix it today/tomorrow 👍

rangoo94: @pavloburchak, I pushed the changes:

* moved the store initialization to `App.tsx` (so it will work in Cloud easily)
* reset store with initialization dependencies

Looks like it works well for me, could you check it?

pavloburchak: @rangoo94 looks ok to me

PR #860, opened 2023-08-25
fix: status page assets

mortada-codes:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/Sider/Sider.tsx
```diff
@@ -95,6 +96,16 @@ const getRoutes = (showSocialLinksInSider: boolean) => [
           active: /environment-management/,
         },
       ]),
+
+  {
```
rangoo94, line 100: Will it be available in OSS?

fivenp: no... but we most likely want to tease it on OSS :)

rangoo94: We should have some page then on this `/status-page` URL, or disable this button when it's not available 👍 Probably it would be best to do it with plugins

src/App.tsx
```diff
@@ -42,11 +38,9 @@ import {PollingIntervals} from '@utils/numbers';
 
 import {MessagePanelWrapper} from './App.styled';
 
-export interface AppProps {
-  plugins: Plugin[];
-}
+export interface AppProps {}
```
haneabogdan, line 41: looks like redundant. can be deleted

haneabogdan (approved):

PR #874, opened 2023-09-13
fix: revert labels improvements

pavloburchak:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


rangoo94:
We should not do that, as it will create a few older bugs again. I will today fix the issues with labels instead.

PR #888, opened 2023-09-27
fix: command args

mortada-codes:
## Changes

- add namespace to the commands if it is different namespace than testkube

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


rangoo94 (approved):
Approving, although:
* `--namespace ${namespace}` should be rather two arguments, so we will avoid problems if sanitizing it later (`["--namespace", namespace]`)
* Should the `--namespace` be passed as argument of the command? I'm not sure if it shouldn't be just after `kubectl testkube`, like `kubectl testkube --namespace custom-name download artifacts`

PR #899, opened 2023-10-09
fix: refetching metrics in entity detail

haneabogdan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/EntityDetails/EntityDetailsLayer.tsx
+++ b/src/components/organisms/EntityDetails/EntityDetailsLayer.tsx
@@ -1,5 +1,5 @@
 import React, {FC, PropsWithChildren, useContext, useEffect, useMemo} from 'react';
-import {useAsync, useInterval} from 'react-use';
+import {useAsync} from 'react-use';
 import useWebSocket from 'react-use-websocket';
 
 import {UseQuery} from '@reduxjs/toolkit/dist/query/react/buildHooks';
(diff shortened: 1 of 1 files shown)


src/components/organisms/EntityDetails/EntityDetailsLayer.tsx
```diff
@@ -206,8 +212,6 @@ const EntityDetailsLayer: FC<PropsWithChildren<EntityDetailsLayerProps>> = ({
     safeRefetch(refetch);
   }, [entity, daysFilterValue]);
 
-  useInterval(() => safeRefetch(refetchMetrics), 2000);
```
rangoo94, line 209: This should rather stay. As long as `skip` condition is truthy, it will rather not make a request. If it's not truthy, we want it to be updated.

src/components/organisms/EntityDetails/EntityDetailsLayer.tsx
```diff
@@ -62,15 +62,21 @@ const EntityDetailsLayer: FC<PropsWithChildren<EntityDetailsLayerProps>> = ({
 
   const {data: rawExecutions, refetch} = useGetExecutions(
     {id, last: daysFilterValue},
-    {pollingInterval: PollingIntervals.long, skip: !isSystemAvailable || (!isClusterAvailable && Boolean(executions))}
+    {
+      pollingInterval: PollingIntervals.long,
+      skip: !isSystemAvailable || (storeDetails?.readOnly && Boolean(executions)),
```
rangoo94, line 67: These should keep the previous conditions, and additionally have another for the expected case. Besides, it's important to take in account case:

* Agent is offline
* User reads the test from API
* Agent becomes online
* User should read the test from API again

With the current condition, it will skip update, as `storeDetails?.readOnly` will be `true`.

rangoo94:
@haneabogdan, this is probably to close, regarding that we want to keep the same behavior for offline results?

PR #902, opened 2023-10-11
fix:  hide abort run when now running tests

mortada-codes:
This PR...

## Changes

- 1 We should safely remove “Abort all executions” if there is NO RUNNING execution. (make sure the it IS visible if there IS an execution running) - ideally this would be covered with an E2E test

-  2 Weird breaking on resize. The select boxes + buttons should be in the same line. The text is currently just cut… We typically use “ellipsis” to have the typo cut by “…” and display the whole title on hover

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/PageBlueprint/PageHeader.tsx
```diff
@@ -24,9 +24,11 @@ const PageHeader: FC<PageHeaderProps> = ({title, pageTitleAddon, description, lo
           {title} <LoadingOutlined />
         </>
       ) : (
-        <PageTitle>
-          {title} {pageTitleAddon}
-        </PageTitle>
+        <Tooltip title={title}>
```
rangoo94, line 27: I'm not sure if a tooltip is a proper option. It may look strange, especially for short titles

mortada-codes: yes, me too. should we enable/disable based on the length of the title?

rangoo94: In the best situation we should show it only when it's truncated. Quick googling shows that most likely it's possible, so it would be great if you could try it: https://stackoverflow.com/a/63214232/2922986

packages/web/src/components/organisms/EntityDetails/EntityDetailsHeader.tsx
```diff
@@ -64,6 +65,18 @@ const EntityDetailsHeader: FC<EntityDetailsHeaderProps> = ({
     });
   };
 
+  const entityOptionsMenu = useMemo(
+    () =>
+      [
+        {key: 1, label: <span onClick={onEditTest}>Edit Test</span>},
```
rangoo94, line 71: `EntityDetailsHeader` is not only used for tests, so the prop should be called `onEdit` and this label should be based on entity type, not static `Edit Test`

rangoo94 (approved):

PR #908, opened 2023-10-12
tech: refactor delete modal

haneabogdan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


packages/web/src/components/molecules/CommonSettings/Delete.tsx
```diff
@@ -14,17 +11,20 @@ interface DeleteProps {
...
-  const {open} = useModal({
+const Delete: FC<DeleteProps> = ({name, description, label, redirectUrl, onDelete}) => {
+  const {open, close} = useModal({
     title: `Delete this ${label}`,
     width: 600,
     content: (
       <DeleteEntityModal
         defaultStackRoute={redirectUrl}
-        useDeleteMutation={useDeleteMutation}
+        onClose={() => {
```
rangoo94, line 24: We could consider using `const {close} = useModal()` from within of `DeleteEntityModal`, but that's optional

packages/web/src/components/molecules/DeleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+        notificationCall('passed', successMessage);
+        onClose();
+        if (onConfirm) {
+          onConfirm();
+        } else if (defaultStackRoute) {
+          back();
+        }
+      })
+      .catch(error => {
+        notificationCall('failed', error.title, error.message);
```
rangoo94, line 67: Is it correct? For requests we use `error.data?.title` and `error.data?.detail`, maybe fallback is needed

packages/web/src/components/molecules/DeleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+  };
+  return (
+    <DeleteModalWrapper>
+      {content || (
+        <FullWidthSpace
+          size={contentSize || 16}
+          direction="vertical"
+          onKeyPress={event => {
+            onEvent(event, onDeleteCallback);
+          }}
```
rangoo94, line 81: Given there is no input - should we just allow Enter key for deletion?

packages/web/src/components/molecules/DeleteEntityModal/DeleteEntityModal.tsx
```diff
@@ -2,101 +2,57 @@ import {useState} from 'react';
...
+            Do you really want to delete this {entityLabel}?
+            <br />
+            All your historical and analytical data will also be removed.
+          </Text>
+          <Text className="regular middle" color={Colors.slate400}>
+            Please enter the name of this {entityLabel} (<span style={{color: Colors.whitePure}}>{name}</span>) to
+            delete it forever.
+          </Text>
+          <Input
+            placeholder={`${capitalize(entityLabel)} name`}
```
rangoo94, line 48: Not important: Could it be disabled during request?

packages/web/src/components/molecules/DeleteModal/DeleteModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+  onConfirm,
+  content,
+  contentSize,
+  title,
+  subtitle,
+  actionText,
+  actionDisabled,
+  defaultStackRoute,
+  successMessage,
+  deleteOnEnter = false,
```
rangoo94, line 46: `deleteOnEnter` is ignored

src/components/molecules/DeleteModal/DeleteModal.tsx
rangoo94, line 1: Probably should be deleted, as it's empty

packages/web/src/components/molecules/CommonSettings/Delete.tsx
haneabogdan: moved to `DeleteEntityModal `

packages/web/src/components/molecules/DeleteModal/DeleteModal.tsx
haneabogdan: adjusted to check for `deleteOnEnter` which is `false` by default.

packages/web/src/components/molecules/DeleteEntityModal/DeleteEntityModal.tsx
haneabogdan: loading state is handled in the child component `DeleteModal`. consider not to do it, because we would have to move that state one level above and handle it for each use case

packages/web/src/components/molecules/DeleteModal/DeleteModal.tsx
haneabogdan: preserved existing logic. looks the same as in other places like `useRunEntity`.

packages/web/src/components/molecules/DeleteEntityModal/DeleteEntityModal.tsx
rangoo94: It doesn't have to be for each specific use case, it can be based on `onDelete` async result

packages/web/src/components/molecules/DeleteModal/DeleteModal.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+        onDelete={onDelete}
+        onClose={onClose}
+        successMessage={successMessage}
+        idToDelete={idToDelete}
+        title="Delete Item"
+        subtitle="Are you sure you want to delete this item?"
+      />
+    );
+
+    expect(getByText('Delete Item')).toBeInTheDocument();
```
rangoo94, line 27: You can use `getByTestId` and compare the text content. Current solution will work too, but I think that `test-id` is more specific. Same for other

packages/web/src/components/molecules/DeleteModal/DeleteModal.tsx
rangoo94: Right, looks like it would be good to check if it's working in `useRunEntity`

packages/web/src/components/molecules/DeleteEntityModal/DeleteEntityModal.tsx
haneabogdan: exactly. that's what we're currently doing inside of `DeleteModal`. what I meant was, if we move this implementation from there, we would have to duplicate it in all the places were using the component.

PR #911, opened 2023-10-18
tech: improve unit test coverage

haneabogdan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/packages/web/src/components/molecules/ArtifactsList/ArtifactsListItem.spec.tsx
+++ b/packages/web/src/components/molecules/ArtifactsList/ArtifactsListItem.spec.tsx
@@ -15,7 +15,7 @@ describe('ArtifactsListItem', () => {
   const testSuiteName = 'test-suite-name';
 
   function createAutoResetSpy(): jest.SpyInstance {
-    const fetchMock = jest.fn();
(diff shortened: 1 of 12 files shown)


packages/web/src/components/molecules/ArtifactsList/ArtifactsListItem.spec.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const artifact = {
+    description: 'artifact-description',
+    name: 'artifact-name',
+    size: 123,
+  };
+  const executionId = '123';
+  const testName = 'test-name';
+  const testSuiteName = 'test-suite-name';
+
+  function createAutoResetSpy(): jest.SpyInstance {
```
rangoo94, line 17: It could return `jest.SpyInstance<typeof fetch>`, so `@ts-ignore` won't be needed. Additionally, initial spy may be something like `jest.fn(() => Promise.reject(new Error('mockImplementation required')))`

PR #913, opened 2023-10-19
fix: executors e2e tests

mortada88:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/packages/e2e-tests/pages/ExecutorsPage.ts
+++ b/packages/e2e-tests/pages/ExecutorsPage.ts
@@ -12,6 +12,6 @@ export class ExecutorsPage {
   }
 
   public async openExecutorSettings(executorName: string): Promise<void> {
-    await this.page.click(`xpath=//div/span[text()="${executorName}"]`);
+    await this.page.click(`div[data-test="${executorName}"]`);
   }
 }
--- a/packages/e2e-tests/pages/ExecutorsPage.ts
+++ b/packages/e2e-tests/pages/ExecutorsPage.ts
@@ -10,4 +10,8 @@ export class ExecutorsPage {
   public async openCreateExecutorDialog(): Promise<void> {
     await this.page.click('xpath=//button//span[contains(text(),"Create a new executor")]'); // TODO: data-test
   }
+
+  public async openExecutorSettings(executorName: string): Promise<void> {
+    await this.page.click(`xpath=//div/span[text()="${executorName}"]`);
+  }
 }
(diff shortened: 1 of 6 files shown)


packages/e2e-tests/pages/ExecutorsPage.ts
```diff
@@ -10,4 +10,8 @@ export class ExecutorsPage {
   public async openCreateExecutorDialog(): Promise<void> {
     await this.page.click('xpath=//button//span[contains(text(),"Create a new executor")]'); // TODO: data-test
   }
+
+  public async openExecutorSettings(executorName: string): Promise<void> {
+    await this.page.click(`xpath=//div/span[text()="${executorName}"]`);
```
tkonieczny, line 15: `data-test` attribute

packages/e2e-tests/tests/executors.spec.ts
```diff
@@ -39,9 +40,55 @@ test(`Create custom container executor`, async ({page}) => {
   await api.removeExecutor(realExecutorName);
 });
 
-test.skip(`Custom container executor - general settings`, async ({page}) => {});
+test(`Custom container executor - general settings`, async ({page}) => {
```
tkonieczny, line 43: I think this case can be extended with setting custom command, and argument

packages/e2e-tests/tests/executors.spec.ts
```diff
@@ -39,9 +40,55 @@ test(`Create custom container executor`, async ({page}) => {
...
+  await mainPage.visitMainPage();
+
+  const navigationSiderPage = new NavigationSiderPage(page);
+  await navigationSiderPage.openMenuItem('executors');
+
+  const executorsPage = new ExecutorsPage(page);
+  await executorsPage.openExecutorSettings(realExecutorName);
+
+  const executorGeneralSettingsPage = new ExecutorGeneralSettingsPage(page);
+  const executorNameInput = await executorGeneralSettingsPage.getExecutorName();
```
tkonieczny, line 59: Usually it's better to user "validators" instead of getting and then comparing the value.
The difference is in this case it's waiting for element matching the selector (`input[id="general-settings-name-type_name"]`), then getting it's value, and then comparing (expect assertion). Element will be checked once, so any state transitions can make the test unstable.

In case of validators it would be looking for the same element, but with specific text value.
So, `executorGeneralSettingsPage.validateExecutorName(executorName)`:
and, all that's needed is checking if the element exists:
```
this.page.locator(`xpath=//input[@id="general-settings-name-type_name" and text="${executorName}"]`)
```


Additionally, on the test level I would use higher-level function instead - like `executorGeneralSettingsPage.validateExecutorDetails` that would then contain `validateExecutorName` on the Page level. The main goal is to keep tests high-level, and keep all of the details on the lower (Page) level.

tkonieczny (changes requested):

PR #896, opened 2023-10-06
fix: adjust visibility and polling logic

haneabogdan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

PR #916, opened 2023-10-24
feat: offline test suites

haneabogdan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/packages/web/src/components/molecules/RunningContext/ContextDisplay.tsx
+++ b/packages/web/src/components/molecules/RunningContext/ContextDisplay.tsx
@@ -1,4 +1,4 @@
-import {FC, ReactNode} from 'react';
+import {FC, MouseEvent, ReactNode, useCallback} from 'react';
 import {To} from 'react-router';
 
 import {Tooltip} from 'antd';
@@ -20,6 +20,15 @@ export interface ContextDisplayProps {
 const ContextDisplay: FC<ContextDisplayProps> = ({url, label, tooltip, wrap, text}) => {
   const isExternal = typeof url === 'string' && hasProtocol(url);
   const navigate = useDashboardNavigate(url || '');
+
+  const onClick = useCallback(
+    (event: MouseEvent) => {
+      event.preventDefault();
+      navigate();
+    },
+    [navigate]
+  );
+
   const content = tooltip ? (
     <Tooltip title={tooltip}>
       <span>{text}</span>
@@ -31,7 +40,9 @@ const ContextDisplay: FC<ContextDisplayProps> = ({url, label, tooltip, wrap, tex
   const item = isExternal ? (
     <ExternalLink href={url}>{content || label}</ExternalLink>
   ) : url ? (
-    <a onClick={navigate}>{content || label}</a>
+    <a href="#" onClick={onClick}>
+      {content || label}
+    </a>
   ) : (
     content || label
   );
--- a/packages/web/src/components/molecules/RunningContext/ContextDisplay.tsx
+++ b/packages/web/src/components/molecules/RunningContext/ContextDisplay.tsx
@@ -31,9 +31,7 @@ const ContextDisplay: FC<ContextDisplayProps> = ({url, label, tooltip, wrap, tex
   const item = isExternal ? (
     <ExternalLink href={url}>{content || label}</ExternalLink>
   ) : url ? (
-    <a href="#" onClick={navigate}>
-      {content || label}
-    </a>
+    <a onClick={navigate}>{content || label}</a>
   ) : (
     content || label
   );
--- a/packages/web/src/components/organisms/EntityDetails/EntityDetailsHeader.tsx
+++ b/packages/web/src/components/organisms/EntityDetails/EntityDetailsHeader.tsx
@@ -71,7 +71,7 @@ const EntityDetailsHeader: FC<EntityDetailsHeaderProps> = ({
       onBack={onBack}
       title={details!.name}
       pageTitleAddon={
-        outOfSync ? (
+        !isAgentAvailable || outOfSync ? (
           <Tag
             title="read-only"
             type="warning"
--- a/packages/web/src/components/organisms/EntityDetails/EntityDetailsHeader.tsx
+++ b/packages/web/src/components/organisms/EntityDetails/EntityDetailsHeader.tsx
@@ -103,8 +103,8 @@ const EntityDetailsHeader: FC<EntityDetailsHeaderProps> = ({
           key="run-now-button"
           type="primary"
           onClick={onRun}
-          disabled={!details || outOfSync}
-          hidden={!mayRun}
+          disabled={!details}
+          hidden={!mayRun || outOfSync}
           loading={isRunning}
         >
           Run now
--- a/packages/web/src/components/organisms/EntityDetails/EntityDetailsHeader.tsx
+++ b/packages/web/src/components/organisms/EntityDetails/EntityDetailsHeader.tsx
@@ -41,6 +41,7 @@ interface EntityDetailsHeaderProps {
   onEditTest: () => void;
   outOfSync?: boolean;
   isAgentAvailable?: boolean;
+  entityLabel: string;
 }
 
 const EntityDetailsHeader: FC<EntityDetailsHeaderProps> = ({
@@ -51,6 +52,7 @@ const EntityDetailsHeader: FC<EntityDetailsHeaderProps> = ({
   onEditTest,
   outOfSync,
   isAgentAvailable,
+  entityLabel,
 }) => {
   const mayRun = usePermission(Permissions.runEntity);
   const {entity, details} = useEntityDetailsPick('entity', 'details');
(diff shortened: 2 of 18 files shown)


packages/web/src/components/molecules/RunningContext/ContextDisplay.tsx
```diff
@@ -31,9 +31,7 @@ const ContextDisplay: FC<ContextDisplayProps> = ({url, label, tooltip, wrap, tex
   const item = isExternal ? (
     <ExternalLink href={url}>{content || label}</ExternalLink>
   ) : url ? (
-    <a href="#" onClick={navigate}>
-      {content || label}
-    </a>
+    <a onClick={navigate}>{content || label}</a>
```
rangoo94, line 34: Shouldn't it stay `href="#"` or become button? Otherwise it won't be looking like a link (as only `a[href]` have pointer cursor).

packages/web/src/components/pages/TestSuites/TestSuiteDetails/TestSuiteDetailsContent.tsx
```diff
@@ -60,34 +62,38 @@ const TestSuiteDetailsContent: FC<TestSuiteDetailsContentProps> = ({tab, setting
...
+                label: 'CLI Commands',
+                children: <CLICommands name={details!.name} bg={Colors.slate800} />,
+              },
+              {
+                key: 'settings',
+                label: 'Settings',
+                children: <TestSuiteSettings active={settingsTab} onChange={setSettingsTab} />,
+              },
+            ].filter(Boolean) as Tab[]
+          }
```
rangoo94, line 96: Either way is fine, but as an alternative you may try without typing:
```suggestion
            [
              {
                key: 'executions',
                label: 'Recent executions',
                children: <RecentExecutionsTab onRun={run} useAbortExecution={useAbortTestSuiteExecutionMutation} />,
              },
              ...(details.readOnly ? [] : {
                key: 'commands',
                label: 'CLI Commands',
                children: <CLICommands name={details!.name} bg={Colors.slate800} />,
              }),
              {
                key: 'settings',
                label: 'Settings',
                children: <TestSuiteSettings active={settingsTab} onChange={setSettingsTab} />,
              },
            ]
          }
```

packages/web/src/components/pages/TestSuites/TestSuiteDetails/TestSuiteSettings/SettingsGeneral/SettingsGeneral.tsx
```diff
@@ -15,17 +13,16 @@ import {useDeleteTestSuiteMutation, useUpdateTestSuiteMutation} from '@services/
...
   const {details} = useEntityDetailsPick('details');
   const deleteTestSuiteExtension = usePluginSlot('deleteTestSuiteExtension');
   const [deleteTestSuite] = useDeleteTestSuiteMutation();
 
   return (
     <>
-      <NameNDescription label="test suite" useUpdateEntity={useUpdateTestSuiteMutation} readOnly={!isWritable} />
-      <Labels label="test suite" useUpdateEntity={useUpdateTestSuiteMutation} readOnly={!isWritable} />
-      {mayDelete && isWritable ? (
+      <NameNDescription label="test suite" useUpdateEntity={useUpdateTestSuiteMutation} readOnly={details.readOnly} />
```
rangoo94, line 23: Not sure if important: Probably it should take in account also the `isWritable` from before - otherwise, there is a gap when the Test Suite is read-only, but it's not perceived as read-only yet

packages/web/src/components/pages/TestSuites/TestSuiteDetails/TestSuiteSettings/SettingsTests/Modals/TestModal.tsx
```diff
@@ -84,7 +86,7 @@ const TestModal: React.FC<TestModalProps> = props => {
         >
           <Select onChange={setTestValue} value={testValue} showSearch placeholder="Select a test..." size="middle">
             {allTestsData.map(item => (
-              <Option value={item.name} key={item.name} title={item.name}>
+              <Option value={item.name} key={item.name}>
```
rangoo94, line 89: This `title` here is used for recognition in E2E tests:
```suggestion
              <Option value={item.name} key={item.name} title={item.name}>
```

packages/web/src/components/pages/TestSuites/TestSuiteDetails/TestSuiteSettings/SettingsVariables/SettingsVariables.tsx
```diff
@@ -1,17 +1,19 @@
...
-  <Variables
-    description="Define environment variables which will be shared across your tests. Variables defined at a Test Suite level will override those defined at a Test level."
-    readOnly={!useSystemAccess(SystemAccess.agent)}
-    useUpdateEntity={useUpdateTestSuiteMutation}
-  />
-);
+import {useEntityDetailsPick} from '@store/entityDetails';
 
+const SettingsVariables: FC = () => {
+  const {details} = useEntityDetailsPick('details');
```
rangoo94, line 10: As in comment somewhere above, we could consider using the `useSystemAccess` to, to avoid moment when the cluster is down, but the UI doesn't have information that the resource is read-only

packages/web/src/components/pages/TestSuites/TestSuitesList/TestSuiteCard.tsx
```diff
@@ -37,7 +38,9 @@ const TestSuiteCard: FC<TestSuiteCardProps> = ({item: {testSuite, latestExecutio
       onAbort={onAbort}
       metrics={metrics}
       dataTest="test-suites-list-item"
-      outOfSync={!isFresh}
+      outOfSync={testSuite.readOnly}
```
rangoo94, line 41: Probably worth to check for `isAgentAvailable` anyway too, as in comments above

packages/web/src/components/pages/TestSuites/TestSuitesList/TestSuitesList.tsx
```diff
@@ -39,7 +41,7 @@ const TestSuitesList: FC = () => {
     isFetching,
   } = useGetTestSuitesQuery(filters || null, {
     pollingInterval: PollingIntervals.everySecond,
-    skip: !isAvailable,
+    skip: !isSystemAvailable || (!isAvailable && Boolean(storeTestSuites)),
```
rangoo94, line 44: Should it avoid querying when there agent is down and some test suites are loaded? I thought that we ended up keeping the polling behavior, in case of i.e. ephemeral environments

packages/web/src/components/pages/TestSuites/TestSuitesList/TestSuitesList.tsx
```diff
@@ -82,7 +84,7 @@ const TestSuitesList: FC = () => {
       queryFilters={filters}
       setQueryFilters={setFilters}
       data={testSuites}
-      isLoading={isLoading}
+      isLoading={isLoading || !isSystemAvailable}
```
rangoo94, line 87: Should it be marked as loading when the system is not available?

packages/web/src/components/pages/TestSuites/TestSuiteDetails/TestSuiteSettings/SettingsGeneral/SettingsGeneral.tsx
haneabogdan: this will not work for mixed case, where the agent is online, but we have a read-only test-suite which is not present on the agent anymore

packages/web/src/components/molecules/RunningContext/ContextDisplay.tsx
haneabogdan: the `<a>` tag preserves it's styling (including pointer cursor) even without `href="#"`.  The reason I it's removed, is because it caused a navigation to the root. Went for the alternative solution of `e.preventDefault`

packages/web/src/components/pages/TestSuites/TestSuitesList/TestSuitesList.tsx
haneabogdan: adjusted the logic [similarly to the tests ](https://github.com/kubeshop/testkube-dashboard/pull/896/files/94ab43bfb5dc97e8e8e805cfe84b22d198398132#r1350120235)

packages/web/src/components/molecules/RunningContext/ContextDisplay.tsx
rangoo94: The `<a>` without `href` behaves differently, as it's constructed differently in the browser (anchor vs link).

Zrzut ekranu 2023-10-27 o 11 30 11

Looks like we are overwriting it with `.ant-typography a` anyway 🙂  I would keep `href` though, to avoid subtle browser behavior changes 👍

packages/web/src/components/organisms/EntityDetails/EntityDetailsHeader.tsx
```diff
@@ -75,8 +77,8 @@ const EntityDetailsHeader: FC<EntityDetailsHeaderProps> = ({
             type="warning"
```
rangoo94, line 77: This tag should be also visible when `!isAgentAvailable`, even when `!outOfSync`

rangoo94 (approved):

PR #926, opened 2023-11-02
Update colors for log parsing

pavloburchak:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


rangoo94 (approved):
I'm not sure if we shouldn't distinguish the background color for the log output (based on ANSI escape codes) and the one for log parsing, but LGTM - if we will have problems because of it, we may adjust that in the future 🙂

PR #934, opened 2023-11-08
fix: gh actions oss integration

mortada-codes:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


rangoo94:
* There should be an option to choose or add the token
* Selected token should be immediately included in the workflow
* `namespace: testkube` could be ignored when it is `testkube`
* It should support the Cloud as well with different workflows
* Other options should be included: `testkube-run-action`, generic Bash script, other CI/CDs

fivenp:
All the things @rangoo94 said.

AND: 

* The whole page refreshes when clicking the CI/CD Tab – this is weird... and we shouldn't
* The editor is in write mode – it should be read only – right? 
* Shouldn't we link to [this article](https://docs.testkube.io/articles/run-tests-with-github-actions) (cc @rangoo94)? 
* On cloud we will have different content – right?

rangoo94:
> Shouldn't we link to [this article](https://docs.testkube.io/articles/run-tests-with-github-actions) (cc @rangoo94)?

This article is regarding the [**testkube-run-action**](https://github.com/kubeshop/testkube-run-action), not the [**testkube-action**](https://github.com/kubeshop/setup-testkube) used in this workflow.

On the other side, I'm considering once again if **testkube-run-action** is needed. It obviously have some advantages over **testkube-action** for running tests, but on the other side it may make less clear why we do have two different actions.

fivenp:
> > Shouldn't we link to [this article](https://docs.testkube.io/articles/run-tests-with-github-actions) (cc @rangoo94)?
> 
> This article is regarding the [**testkube-run-action**](https://github.com/kubeshop/testkube-run-action), not the [**testkube-action**](https://github.com/kubeshop/setup-testkube) used in this workflow.
> 
> On the other side, I'm considering once again if **testkube-run-action** is needed. It obviously has some advantages over **testkube-action** for running tests, but on the other side, it may make it less clear why we do have two different actions.

Agree... then let's stick to using the `testkube-action` everywhere :)

rangoo94:
Closing in favor of https://github.com/kubeshop/testkube-cloud-ui/pull/512 and accompanying one in OSS

PR #940, opened 2023-11-15
fix: catch validate fields so app doesn't crash TKC-863

topliceanurazvan:
## Changes

- Handle validateFields and don't crash the app on error

## Fixes

- https://linear.app/kubeshop/issue/TKC-863/%5Bui%5D-creating-triggerwebhook-with-empty-required-valued-would-crash

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


rangoo94 (approved):

rangoo94:
I'm not sure if the error should be ignored, but looks rather OK 👍

PR #942, opened 2023-11-16
refactor: filter tests/testSuites based on action TKC-871

topliceanurazvan:
## Changes

- Filter resources based on the action ( `run test` or `run testsuite` )
- Searchable resource selector
- Searchable K8s resource selector from conditions

## Fixes

- [**TKC-871:** Trigger action not filtering corresponding resources](https://linear.app/kubeshop/issue/TKC-871/%5Bui%5D-trigger-action-not-filtering-corresponding-resources)

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/packages/web/src/components/organisms/TriggersFormItems/ActionFormItems.tsx
+++ b/packages/web/src/components/organisms/TriggersFormItems/ActionFormItems.tsx
@@ -14,11 +14,11 @@ import TriggerSelectorSwitcher from './TriggerSelectorSwitcher';
 const ActionFormItems = () => {
   const {keyMap} = useTriggersPick('keyMap');
 
-  const [actionValue, setActionValue] = useState<string>('');
   const [switcherValue, setSwitcherValue] = useState('label');
 
   const form = Form.useFormInstance();
 
+  const actionValue = Form.useWatch('action', form);
   const nameSelector = form.getFieldValue('testNameSelector');
 
   useEffect(() => {
@@ -47,12 +47,7 @@ const ActionFormItems = () => {
         name="action"
         rules={[required]}
       >
-        <Select
-          value={actionValue}
-          onChange={value => setActionValue(value)}
-          options={actionOptions}
-          placeholder="Select a testkube related action"
-        />
+        <Select options={actionOptions} placeholder="Select a testkube related action" />
       </Form.Item>
       <Space size={16} direction="vertical" style={{width: '100%'}}>
         <TriggerSelectorSwitcher
@@ -79,8 +74,8 @@ const ActionFormItems = () => {
             rules={[required]}
           >
             <ResourceTriggerSelect
-              $allowTests={actionValue === 'run test'}
-              $allowTestSuites={actionValue === 'run testsuite'}
+              allowTests={actionValue === 'run test'}
+              allowTestSuites={actionValue === 'run testsuite'}
             />
           </Form.Item>
         )}
--- a/packages/web/src/components/organisms/TriggersFormItems/ActionFormItems.tsx
+++ b/packages/web/src/components/organisms/TriggersFormItems/ActionFormItems.tsx
@@ -17,11 +17,17 @@ const ActionFormItems = () => {
   const [actionValue, setActionValue] = useState<string>('');
   const [switcherValue, setSwitcherValue] = useState('label');
 
-  const nameSelector = Form.useFormInstance().getFieldValue('testNameSelector');
+  const form = Form.useFormInstance();
+
+  const nameSelector = form.getFieldValue('testNameSelector');
 
   useEffect(() => {
     setSwitcherValue(nameSelector ? 'name' : 'label');
   }, [nameSelector]);
+
+  useEffect(() => {
+    form.resetFields(['testNameSelector']);
+  }, [actionValue]);
 
   const actionOptions = keyMap?.actions
     .map((actionItem: string) =>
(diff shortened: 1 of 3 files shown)


fivenp:
me likes :) 

shall we also include this select? 
image

packages/web/src/components/organisms/TriggersFormItems/ActionFormItems.tsx
```diff
@@ -66,7 +72,7 @@ const ActionFormItems = () => {
             name="testNameSelector"
             rules={[required]}
           >
-            <ResourceTriggerSelect />
+            <ResourceTriggerSelect actionValue={actionValue} />
```
rangoo94, line 75: Probably it would be better to move the responsibility - `ActionFormItems` understands the action value, but `ResourceTriggerSelect` shouldn't.

Instead, you may add props `tests`/`testSuites` (or `allowTests`/`allowTestSuites`) to the `ResourceTriggerSelect`:
```suggestion
            <ResourceTriggerSelect allowTests={actionValue === 'run test'} allowTestSuites={actionValue === 'run testsuite'} />
```

This way:
* `ResourceTriggerSelect` may easily skip unnecessary resources query
* `ResourceTriggerSelect` may show any set of resources, as per parent's expectation

packages/web/src/components/organisms/TriggersFormItems/ResourceTriggerSelect/ResourceTriggerSelect.tsx
```diff
@@ -24,38 +24,46 @@ import {StyledResourceOptionWrapper} from './ResourceTriggerSelect.styled';
...
+const ResourceTriggerSelect: FC<ResourceTriggerSelectProps> = props => {
+  const {$allowTestSuites, $allowTests, disabled, value, onChange, ...rest} = props;
+
   const isClusterAvailable = useSystemAccess(SystemAccess.agent);
 
   const {executors = []} = useExecutorsPick('executors');
 
-  const {data: tests = []} = useGetAllTestsQuery(null, {skip: !isClusterAvailable});
+  const {data: tests = []} = useGetAllTestsQuery(null, {skip: !isClusterAvailable || !$allowTests});
   const {data: testSuites = []} = useGetAllTestSuitesQuery(null, {
```
rangoo94, line 42: One bug that was already before - `tests = []` and `testSuites = []` should be avoided, as both `tests` and `testSuites` are used later for `useMemo` dependencies. Having `tests = []` here, will always result in recomputing, as `[]` will always be a new array.

packages/web/src/components/organisms/TriggersFormItems/ResourceTriggerSelect/ResourceTriggerSelect.tsx
```diff
@@ -24,38 +24,46 @@ import {StyledResourceOptionWrapper} from './ResourceTriggerSelect.styled';
 const {Option, OptGroup} = Select;
 
 interface ResourceTriggerSelectProps {
+  $allowTests: boolean;
+  $allowTestSuites: boolean;
```
rangoo94, line 28: Better to keep them without `$`. We use `$` prefix only for styled-component's [**transient props**](https://styled-components.com/docs/api#transient-props), not for logical props.

packages/web/src/components/organisms/TriggersFormItems/ActionFormItems.tsx
```diff
@@ -14,14 +14,21 @@ import TriggerSelectorSwitcher from './TriggerSelectorSwitcher';
 const ActionFormItems = () => {
   const {keyMap} = useTriggersPick('keyMap');
 
+  const [actionValue, setActionValue] = useState<string>('');
```
rangoo94, line 17: You don't need to store `actionValue` in state, instead you may use `Form.useWatch`:
```suggestion
  const actionValue = Form.useWatch('action');
```
or optionally `form.getFieldValue('action')`

packages/web/src/components/organisms/TriggersFormItems/ResourceTriggerSelect/ResourceTriggerSelect.tsx
topliceanurazvan: The only thing is that React is complaining about using `allowTests` and that we should use `allowtests`... Not sure how to skip this warning

rangoo94: I see, the problem is in line 66:
```tsx
    <Select optionLabelProp="key" placeholder="Your testkube resource" showSearch {...props}>
```

The `Select` gets all the `props` down, while it should only have `rest`:

```diff
-     <Select optionLabelProp="key" placeholder="Your testkube resource" showSearch {...props}>
+     <Select optionLabelProp="key" placeholder="Your testkube resource" showSearch {...rest}>
```

Otherwise, props like `allowTests` are passed to `Select` and not recognized.

packages/web/src/components/organisms/TriggersFormItems/ActionFormItems.tsx
```diff
@@ -16,12 +16,19 @@ const ActionFormItems = () => {
 
   const [switcherValue, setSwitcherValue] = useState('label');
 
-  const nameSelector = Form.useFormInstance().getFieldValue('testNameSelector');
+  const form = Form.useFormInstance();
+
+  const actionValue = Form.useWatch('action', form);
+  const nameSelector = form.getFieldValue('testNameSelector');
```
rangoo94, line 22: Optionally `useWatch` could be used there too (to use subscription instead of one time value), but not important 🙂

rangoo94 (approved):

PR #954, opened 2023-11-22
CLI Commands for Test & TestSuite update

devcatalin:
This PR...

## Changes

Closes the following issue: https://github.com/kubeshop/testkube/issues/1617

## Notes

- We have a `literalToColor` map that defines colors for each literal, and the color for the `update` literal within the commands is set to `amber300` , which is different from the color that `get` and `delete` have. Not sure if we want to align the colors as it might break the design in other places? @fivenp 

- The `update` commands require passing `--name`, they do not work like the other commands.

## Screenshots

image

image

### DIFF
--- a/packages/web/src/components/molecules/CLICommands/CLICommands.tsx
+++ b/packages/web/src/components/molecules/CLICommands/CLICommands.tsx
@@ -16,7 +16,12 @@ import {useTelemetry} from '@telemetry/hooks';
 
 import CopyCommand from './CopyCommand';
 
-type CLIScriptModifier = 'isFinished' | 'canHaveArtifacts' | 'isRunPermission' | 'isDeletePermission';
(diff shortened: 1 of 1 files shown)


rangoo94 (approved):
The code LGTM, I'm just not sure if we should have it; that `update` command is void, and the `testkube update *` is tricky too (due to partial updates)

haneabogdan (approved):

fivenp:
@devcatalin – we initially planned to have different colors for the action - so I think it would be great to update get + delete colors too :) 
https://www.figma.com/file/asfUPepLHtyutWWR0pNGfK/Testkube-V2?type=design&node-id=824-37812&mode=design&t=HovPc7YCyA064pfE-4

That said we should make sure that we only change get + delete colors as you said and not impact others

PR #982, opened 2024-01-04
fix: escape arguments containing spaces

topliceanurazvan:
## Fixes
 
- Escape arguments in textarea with spaces

## How to test it

- Add arguments to a test that contain space characters

## screenshots

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
+++ b/packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
@@ -48,17 +48,21 @@ const Arguments: React.FC<ArgumentsProps> = ({readOnly}) => {
   const initialArgs = useMemo(() => entityArgs?.join('\n') || '', [entityArgs]);
 
   const currentArgs = Form.useWatch('args', form) || '';
-  const currentArgsArray =
-    currentArgs
-      .split('\n')
-      .map(arg => (arg.includes(' ') ? `"${arg}"` : arg))
-      .filter(Boolean) || [];
+
+  const formattedArgs = useMemo(() => {
+    const currentArgsArray = currentArgs.split('\n').filter(Boolean);
+    return currentArgsArray.map(arg => {
+      const trimmedArg = arg.replace(/\s+/g, ' ').trim();
+
+      return arg.includes(' ') ? `"${trimmedArg}"` : trimmedArg;
+    });
+  }, [currentArgs]);
 
   const isPrettified = useMemo(() => currentArgs === prettifyArguments(currentArgs), [currentArgs]);
 
   const onSaveForm = async () => {
     // Reset the form when there is no actual change
-    if (currentArgsArray.join('\n') === initialArgs) {
+    if (formattedArgs.join('\n') === initialArgs) {
       form.resetFields();
     }
 
@@ -66,7 +70,7 @@ const Arguments: React.FC<ArgumentsProps> = ({readOnly}) => {
       ...details,
       executionRequest: {
         ...details.executionRequest,
-        args: currentArgsArray,
+        args: formattedArgs,
       },
     };
 
@@ -104,7 +108,7 @@ const Arguments: React.FC<ArgumentsProps> = ({readOnly}) => {
       onCancel={onCancel}
     >
       <ArgumentsWrapper>
-        <CopyCommand command={currentArgsArray.join(' ')} isBordered additionalPrefix="executor-binary" />
+        <CopyCommand command={formattedArgs.join(' ')} isBordered additionalPrefix="executor-binary" />
         <FullWidthSpace size={16} direction="vertical">
           <Text className="regular middle" color={Colors.slate400}>
             Arguments passed to the executor (concat and passed directly to the executor)
--- a/packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
+++ b/packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
@@ -48,15 +48,17 @@ const Arguments: React.FC<ArgumentsProps> = ({readOnly}) => {
   const initialArgs = useMemo(() => entityArgs?.join('\n') || '', [entityArgs]);
 
   const currentArgs = Form.useWatch('args', form) || '';
-  const currentArgsInline = currentArgs.replace(/\s+/g, ' ').trim();
+  const currentArgsArray =
+    currentArgs
+      .split('\n')
+      .map(arg => (arg.includes(' ') ? `"${arg}"` : arg))
+      .filter(Boolean) || [];
+
   const isPrettified = useMemo(() => currentArgs === prettifyArguments(currentArgs), [currentArgs]);
 
-  const onSaveForm = () => {
-    const values = form.getFieldsValue();
-    const argVal = values.args?.trim().split('\n').filter(Boolean) || [];
-
+  const onSaveForm = async () => {
     // Reset the form when there is no actual change
-    if (argVal.join('\n') === initialArgs) {
+    if (currentArgsArray.join('\n') === initialArgs) {
       form.resetFields();
     }
 
@@ -64,18 +66,17 @@ const Arguments: React.FC<ArgumentsProps> = ({readOnly}) => {
       ...details,
       executionRequest: {
         ...details.executionRequest,
-        args: argVal,
+        args: currentArgsArray,
       },
     };
 
-    return updateTest({
+    const res = await updateTest({
       id: details.name,
       data: successRecord,
-    })
-      .then(displayDefaultNotificationFlow)
-      .then(() => {
-        notificationCall('passed', 'Variables were successfully updated.');
-      });
+    });
+
+    displayDefaultNotificationFlow(res);
+    notificationCall('passed', 'Variables were successfully updated.');
   };
 
   const onPrettify = useLastCallback(() => form.setFieldValue('args', prettifyArguments(currentArgs)));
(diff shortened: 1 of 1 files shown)


packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -28,6 +28,14 @@ import {prettifyArguments} from '@utils/prettifyArguments';
 
 import {ArgumentsWrapper} from './Arguments.styled';
 
+const formatArgsArray = (args: string[]) => {
+  return args.map(arg => {
+    const trimmedArg = arg.replace(/\s+/g, ' ').trim();
```
rangoo94, line 33: It should rather not trim the argument, as the argument may start from spaces (rather edge case, but still):
```yaml
args:
- "   i want to have spaced argument  "
```

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -28,6 +28,14 @@ import {prettifyArguments} from '@utils/prettifyArguments';
...
+
+    return arg.includes(' ') && !/^(['"]).*\1$/.test(arg) ? `"${trimmedArg}"` : trimmedArg;
```
rangoo94, line 35: Checking the starting character won't work, as:
```yaml
args:
- "some value"
```
Will result in:
```json
["some value"]
```

It will start from `['"]` only in case it specifically has it, and it should be persisted:
```yaml
args:
- "\"abc"
```

Additionally, the quotes inside needs to be escaped too. As an example `abra"kadabra` should not be escaped as `"abra"kadabra"`, but rather `"abra\"kadabra"`

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -28,6 +28,19 @@ import {prettifyArguments} from '@utils/prettifyArguments';
 
 import {ArgumentsWrapper} from './Arguments.styled';
 
+const formatArgsArray = (args: string[]) => {
+  return args.map(arg => {
+    const trimmedArg = arg.replace(/\s+|(?<=[^'])['"](?=[^'])/g, match => {
+      if (match.trim() === '') {
+        return ' '; // Replaces multiple spaces with a single space
```
rangoo94, line 35: It should not replace multiple spaces in a context of single argument.

It would replace:
```yaml
# abc "   def  xyz   " "abc"
command:
  - abc
args:
  - "   def  xyz   "
  - "abc"
```

With:
```yaml
# abc " def xyz " "abc"
command:
  - abc
args:
  - " def xyz "
  - "abc"
```

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -28,6 +28,19 @@ import {prettifyArguments} from '@utils/prettifyArguments';
...
+      }
+      return `\\${match}`; // Escapes quotes
```
rangoo94, line 37: It would work, although won't display nice. It will unnecessarily escape some characters.

As an example `"abc'def"` will become `"abc\'def"` despite, `\` is redundant here

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -28,6 +28,10 @@ import {prettifyArguments} from '@utils/prettifyArguments';
 
 import {ArgumentsWrapper} from './Arguments.styled';
 
+const formatArgsArray = (args: string[]) => {
+  return args.map(arg => (arg.includes(' ') ? `"${arg}"` : arg));
```
rangoo94, line 32: Missing escaping of `"` inside.

```
args:
- 'abc" def'
```
will become invalid `"abc" def"` instead of `"abc\" def"`

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -103,7 +105,7 @@ const Arguments: React.FC<ArgumentsProps> = ({readOnly}) => {
       onCancel={onCancel}
     >
       <ArgumentsWrapper>
-        <CopyCommand command={currentArgsInline} isBordered additionalPrefix="executor-binary" />
+        <CopyCommand command={currentArgs} isBordered additionalPrefix="executor-binary" />
```
rangoo94, line 108: Won't it make it multiline? Maybe sth like:

```
// Above
const prettifiedArgs = useMemo(() => prettifyArguments(currentArgs), [currentArgs]); // and replace `isPrettified` to not use `useMemo`
const currentArgsInline = useMemo(() => prettifiedArgs.replace(/\n+/g, ' '), [prettifiedArgs]);
```

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -28,6 +28,10 @@ import {prettifyArguments} from '@utils/prettifyArguments';
 
 import {ArgumentsWrapper} from './Arguments.styled';
 
+const formatArgsArray = (args: string[]) => {
```
rangoo94, line 31: Could be worth to extract it as some `escapeArgument` function along with unit tests 👍

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
rangoo94: Additionally, it sould escape quotes even when there is no space, otherwise:
```
args:
- 'abc"def'
```

will become `abc"def` instead of `abc\"def`

rangoo94: Plus, it should check for any kind of space `/\s/.test(arg)`

rangoo94: We could as well just use external library for quoting, like [**shell-quote**](https://www.npmjs.com/package/shell-quote)

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -18,6 +18,8 @@ import {Permissions, usePermission} from '@permissions/base';
 
 import {useUpdateTestMutation} from '@services/tests';
 
+import {escapeArguments} from '@src/utils/escapeArguments';
```
rangoo94, line 21:
```suggestion
import {escapeArguments} from '@utils/escapeArguments';
```

packages/web/src/utils/escapeArguments.ts
```diff
@@ -0,0 +1,7 @@
+export const escapeArguments = (args: string[]) => {
+  return args.map(arg => {
+    const escapedArg = arg.replace(/(")/g, '\\$1');
```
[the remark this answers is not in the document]
rangoo94: Still quoting of `'` missing (and not important: maybe backtick, new line), i.e. for `args: ["abc'def"]` it will output `abc'def` instead of `"abc'def"` or `abc\'def`

Maybe the simplest fix would be to change regex below to quote if `[\s']` is matching

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -45,15 +46,18 @@ const Arguments: React.FC<ArgumentsProps> = ({readOnly}) => {
...
+  const initialArgs = useMemo(() => escapeArguments(entityArgs)?.join('\n') || '', [entityArgs]);
 
   const currentArgs = Form.useWatch('args', form) || '';
-  const currentArgsInline = currentArgs.replace(/\s+/g, ' ').trim();
-  const isPrettified = useMemo(() => currentArgs === prettifyArguments(currentArgs), [currentArgs]);
 
-  const onSaveForm = () => {
-    const values = form.getFieldsValue();
-    const argVal = values.args?.trim().split('\n').filter(Boolean) || [];
+  // Above
```
rangoo94, line 53:
```suggestion
```

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -45,15 +46,18 @@ const Arguments: React.FC<ArgumentsProps> = ({readOnly}) => {
...
+  const prettifiedArgs = useMemo(() => prettifyArguments(currentArgs), [currentArgs]); // and replace `isPrettified` to not use `useMemo`
```
rangoo94, line 54:
```suggestion
  const prettifiedArgs = useMemo(() => prettifyArguments(currentArgs), [currentArgs]);
```

packages/web/src/components/pages/Tests/TestDetails/TestSettings/SettingsVariables/Arguments.tsx
```diff
@@ -45,15 +46,18 @@ const Arguments: React.FC<ArgumentsProps> = ({readOnly}) => {
...
+  const currentArgsInline = useMemo(() => prettifiedArgs.replace(/\n+/g, ' '), [prettifiedArgs]);
+
+  const isPrettified = Boolean(currentArgs === prettifyArguments(currentArgs));
```
rangoo94, line 57: Unnecessary boolean cast, unnecessary `prettifyArguments` call:
```suggestion
  const isPrettified = currentArgs === prettifiedArgs;
```

packages/web/src/utils/escapeArguments.ts
```diff
@@ -0,0 +1,7 @@
+export const escapeArguments = (args: string[]) => {
+  return args.map(arg => {
+    const escapedArg = arg.replace(/(["'`])/g, '\\$1');
+
+    return /\s/.test(escapedArg) ? `"${escapedArg}"` : escapedArg;
```
rangoo94, line 5: `abc' def` will result in `"abc\' def"` (unnecessary escaping of `'`)

```suggestion
    return /\s/.test(escapedArg) ? `"${escapedArg.replace(/\\'/g, '\''}"` : escapedArg;
```
+ unit test

rangoo94 (approved):
LGTM 👍

PR #984, opened 2024-01-08
feat: on log click

topliceanurazvan:
## Changes

- If `L` search param is found, scroll to the line and select log line

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/packages/web/src/components/molecules/Console/Console.tsx
+++ b/packages/web/src/components/molecules/Console/Console.tsx
@@ -10,15 +10,13 @@ import {
   useRef,
   useState,
 } from 'react';
-import {useCopyToClipboard, usePrevious, useUpdate} from 'react-use';
+import {usePrevious, useUpdate} from 'react-use';
 
 import {escapeCarriageReturn} from 'escape-carriage';
 import {debounce} from 'lodash';
 
 import {useEventCallback} from '@hooks/useEventCallback';
 import {useLastCallback} from '@hooks/useLastCallback';
-
-import {notificationCall} from '@molecules';
 
 import * as S from './Console.styled';
 import {ConsoleLine} from './ConsoleLine';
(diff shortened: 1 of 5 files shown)


rangoo94:
Should it copy a log line on click? It may cause problems for Users who just click due to "nervous tick", while reading the logs. Cc: @fivenp

fivenp:
agree with dawid... it should not copy on click - but mark the line on click (and adapt the url for the permalink)

pretty much exactly like github does but without multiline (unless its easy to do)

packages/web/src/components/molecules/Console/Console.tsx
```diff
@@ -43,141 +47,186 @@ export interface ConsoleRef {
...
-      get container() {
-        return containerRef.current;
-      },
-      isScrolledToStart() {
-        return containerRef.current?.scrollTop === 0;
-      },
-      scrollToStart: () => {
-        if (containerRef.current) {
-          containerRef.current.scrollTo(0, 0);
+export const Console = forwardRef<ConsoleRef, ConsoleProps>(
```
rangoo94, line 50: `isRunning` could be handled outside of the`Console` - it shouldn't disallow selecting the line when `!isRunning`, the external handler should decide what to do with the selected line

packages/web/src/components/molecules/Console/ConsoleLines.tsx
```diff
@@ -9,11 +9,23 @@ export const ConsoleLines: FC<{
   start: number;
   maxDigits: number;
   LineComponent: ConsoleProps['LineComponent'];
-}> = memo(({lines, start, maxDigits, LineComponent = ConsoleLine}) => (
+  onSelectLine?: (line: number) => void;
```
rangoo94, line 12: It's `onClick` actually, the line shouldn't necessary now that it selecting, but knows that it's clicked

packages/web/src/components/molecules/Console/ConsoleLines.tsx
```diff
@@ -9,11 +9,23 @@ export const ConsoleLines: FC<{
...
+  selectedLine?: number;
```
rangoo94, line 13: I'm not sure if there should be `selectedLine` at all, but if so, it should be `selected: boolean` instead

packages/web/src/components/molecules/Console/Console.tsx
topliceanurazvan: We still need the `isRunning` boolean for scrolling to line number on render, as when the exeucution is ongoing we shouldn't scroll to the found query param

packages/web/src/components/molecules/Console/ConsoleLines.tsx
topliceanurazvan: We can't really have `selected`, since in this component is about multiple lines, and we would need to check the selected line with the rendered one 🤔

topliceanurazvan: Update: after latest commit this prop has been removed

packages/web/src/components/molecules/Console/Console.tsx
topliceanurazvan: Update: after latest commit, `isRunning` was removed and not used

devcatalin (approved):
LGTM

PR #994, opened 2024-01-22
fix: logs rendering when switching drawer tabs

topliceanurazvan:
## Fixes

- Render logs when switching between tabs

## screenshots

https://github.com/kubeshop/testkube-dashboard/assets/47887589/b97cdeb8-720a-41d1-969d-25268525b97f

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/packages/web/src/components/molecules/Console/Console.tsx
+++ b/packages/web/src/components/molecules/Console/Console.tsx
@@ -9,7 +9,7 @@ import {
   useRef,
   useState,
 } from 'react';
-import {useSearchParams} from 'react-router-dom';
+import {useParams, useSearchParams} from 'react-router-dom';
 import {usePrevious, useUpdate} from 'react-use';
 
 import {escapeCarriageReturn} from 'escape-carriage';
@@ -48,6 +48,8 @@ export const Console = forwardRef<ConsoleRef, ConsoleProps>(({content, wrap, Lin
   const processor = useMemo(() => LogProcessor.from(escapeCarriageReturn(content)), [content]);
   const containerRef = useRef<HTMLDivElement | null>(null);
   const maxDigits = processor.maxDigits;
+
+  const {execDetailsTab} = useParams();
 
   const rerender = useUpdate();
 
@@ -193,7 +195,7 @@ export const Console = forwardRef<ConsoleRef, ConsoleProps>(({content, wrap, Lin
   }, [viewportStart, viewportEnd, scrollTop, clientHeight, total]);
 
   // Re-render on scroll
-  const rerenderDebounce = useMemo(() => debounce(rerender, 5), []);
+  const rerenderDebounce = useMemo(() => debounce(rerender, 5), [rerender]);
   const onScroll = () => {
     const viewport = getViewport();
     if (viewport.start !== viewportStart || viewport.end !== viewportEnd) {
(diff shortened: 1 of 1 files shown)


haneabogdan (approved):

packages/web/src/components/molecules/Console/Console.tsx
```diff
@@ -49,6 +49,8 @@ export const Console = forwardRef<ConsoleRef, ConsoleProps>(({content, wrap, Lin
   const containerRef = useRef<HTMLDivElement | null>(null);
   const maxDigits = processor.maxDigits;
 
+  const {execDetailsTab} = useParams();
```
rangoo94, line 52: Do we actually need that? It would be best if we won't check for params from the console

packages/web/src/components/molecules/Console/Console.tsx
```diff
@@ -193,7 +195,7 @@ export const Console = forwardRef<ConsoleRef, ConsoleProps>(({content, wrap, Lin
   }, [viewportStart, viewportEnd, scrollTop, clientHeight, total]);
 
   // Re-render on scroll
-  const rerenderDebounce = useMemo(() => debounce(rerender, 5), []);
+  const rerenderDebounce = useMemo(() => debounce(rerender, 5), [rerender]);
```
rangoo94, line 198: Does `rerender` persist the identity, or the `rerenderDebounce` is recalculated every render?

packages/web/src/components/molecules/Console/Console.tsx
topliceanurazvan: It seems like the logs are not being rendered when switching between tabs... so I used the active tab from params just to rerender when switching back to the logs output

rangoo94: Could we base on i.e. visibility instead? It would be more flexible

packages/web/src/components/molecules/Console/Console.tsx
topliceanurazvan: It persist and it is not recalculated on every rerender

packages/web/src/components/molecules/Console/Console.tsx
topliceanurazvan: Pushed change

packages/web/src/components/molecules/Console/Console.tsx
```diff
@@ -202,6 +202,29 @@ export const Console = forwardRef<ConsoleRef, ConsoleProps>(({content, wrap, Lin
   };
   useEventCallback('scroll', onScroll, containerRef?.current);
 
+  useEffect(() => {
```
rangoo94, line 205: Could be [`useClientRect`](https://github.com/kubeshop/testkube-dashboard/blob/develop/packages/web/src/hooks/useClientRect.ts), [`useInViewport`](https://github.com/kubeshop/testkube-dashboard/blob/develop/packages/web/src/hooks/useInViewport.ts) or directly [`react-use/useIntersection`](https://github.com/streamich/react-use/blob/master/src/useIntersection.ts)

rangoo94 (approved):

packages/web/src/components/molecules/Console/Console.tsx
topliceanurazvan: Refactored

PR #1014, opened 2024-02-12
fix: [TKC-1439] download artifact actual content

topliceanurazvan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/packages/web/src/services/artifacts.ts
+++ b/packages/web/src/services/artifacts.ts
@@ -2,15 +2,45 @@ import {getApiEndpoint} from '@services/apiEndpoint';
 
 import {getRtkBaseUrl, getRtkIdToken} from '@utils/rtk';
 
-const downloadFile = async (blob: Blob, fileName: string): Promise<void> => {
-  const url = URL.createObjectURL(blob);
-  const a = document.createElement('a');
-  a.href = url;
-  a.download = fileName;
-  document.body.appendChild(a);
-  a.click();
-  document.body.removeChild(a);
-  URL.revokeObjectURL(url);
+export const fetchContentFromSignedURL = async (signedURL: string): Promise<Blob> => {
+  const response = await fetch(signedURL);
+
+  if (!response.ok) {
+    throw new Error(`Failed to fetch content from signed URL: ${response.status} ${response.statusText}`);
+  }
+
+  return response.blob();
+};
+
+export const getResponseFileUrl = async (response: Response) => {
+  if (!response.status || response.status >= 300) {
+    throw new Error('Something went wrong');
+  }
+
+  if (response.headers.get('content-type') === 'application/json') {
+    const data = await response.json();
+    const fileUrl = data.data?.url || data.url;
+    return fileUrl;
+  }
+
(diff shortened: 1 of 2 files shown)


packages/web/src/services/artifacts.ts
```diff
@@ -2,15 +2,45 @@ import {getApiEndpoint} from '@services/apiEndpoint';
...
+  try {
+    const content = await fetchContentFromSignedURL(signedURL);
+    const url = URL.createObjectURL(content);
+    const a = document.createElement('a');
+    a.href = url;
+    a.download = fileName;
+    document.body.appendChild(a);
+    a.click();
+    document.body.removeChild(a);
+    URL.revokeObjectURL(url);
```
rangoo94, line 40: This won't revoke object URL created in `getResponseFileUrl`; likely instead the `downloadFile` should always receive a `response: Response` as parameter instead of `signedURL: string`

rangoo94: Or, you may just revert to the previous implementation

topliceanurazvan: I've made changes to fix that, unfortunately reverting was not an option as previously the download was not always happening and it was previewing the file instead

packages/web/src/services/artifacts.ts
```diff
@@ -2,15 +2,47 @@ import {getApiEndpoint} from '@services/apiEndpoint';
...
+  }
+
+  const fileUrl = window.URL.createObjectURL(await response.blob());
+  await downloadFile(fileUrl, fileName);
+  URL.revokeObjectURL(fileUrl);
+};
+
+const downloadFile = async (signedURL: string, fileName: string): Promise<void> => {
+  try {
+    const content = await fetchContentFromSignedURL(signedURL);
```
rangoo94, line 34: It would be much better to avoid the second fetch, as the User won't see the download progress bar otherwise, but it's minor at the moment - we will have soon a new BE endpoint, so we will be able to simplify that logic anyway - we won't need to check the JSON file for signed URL.

rangoo94 (approved):

packages/web/src/services/artifacts.ts
rangoo94: I don't think it's possible that it was not always happening - https://github.com/kubeshop/testkube-dashboard/blob/75af794f0dcb14fe7a13e026e891e7417a5e4f42/packages/web/src/services/artifacts.ts should definitely work

topliceanurazvan: As far as I tested it on different file types, in some cases it was not downloading, but rather showing the content in the browser... 😕

rangoo94: It was using `a[download]`, so it is forcing download - the browser won't ever display the file (even HTML) if it's triggered this way

topliceanurazvan: Well that I'm not sure what was the problem as this was the flow on my machine at least...

PR #1018, opened 2024-02-19
feat: Plugin Overlay for static, high priority, immutable configuration

devcatalin:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


packages/web/src/plugins/rtk/plugin.tsx
```diff
@@ -13,37 +15,33 @@ export interface RtkService {
...
+      middleware: getDefaultMiddleware => [
+        ...getDefaultMiddleware(),
+        createLogger({
+          predicate: (_, action) => {
+            return (
+              action.type.startsWith('testsApi/executeQuery') || action.type.startsWith('testSuitesApi/executeQuery')
+            );
+          },
+          collapsed: true,
         }),
```
rangoo94, line 35: Likely it's some debugging
```suggestion
```

packages/web/src/plugins/tests-and-test-suites/plugin.tsx
```diff
@@ -142,8 +140,8 @@ export default createPlugin('oss/tests-and-test-suites')
   .data({useLogOutput, useLogOutputPick, useLogOutputField, useLogOutputSync})
 
   .init(tk => {
-    tk.slots.rtkServices.add(testSuitesApi);
-    tk.slots.rtkServices.add(testsApi);
+    // tk.slots.rtkServices.add(testSuitesApi);
+    // tk.slots.rtkServices.add(testsApi);
```
rangoo94, line 145:
```suggestion
```

packages/web/src/plugins/cluster-status/plugin.tsx
```diff
@@ -38,6 +36,6 @@ export default createPlugin('dashboard/cluster-status')
     }, [apiEndpoint]);
   })
 
-  .init(tk => {
-    tk.slots.rtkServices.add(configApi);
-  });
+  .init();
+
+RtkPlugin.overlay.appendContext({configApi});
```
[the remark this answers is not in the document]
devcatalin: we didn't have any use cases for that currently, but if we need it at some point, we can try creating an abstraction over the [injectEndpoints](https://redux-toolkit.js.org/rtk-query/usage/code-splitting) method from RTK which is the ideal way of including APIs conditionally 🤔 other options would be looking at `replaceReducer` or `rootReducer.inject` , but we have to create the store only once, otherwise we run into problems

PR #1029, opened 2024-04-03
refactor: add nginx security headers

devcatalin:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/packages/web/scripts/security.sh
+++ b/packages/web/scripts/security.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+API_HOST=$(echo $REACT_APP_API_SERVER_ENDPOINT | sed -e 's|http://||g' -e 's|https://||g')
+
 tempFile=$(mktemp /etc/nginx/tempfile.XXXXXXXX)
 
 cat > "${tempFile}" <<EOF
@@ -7,7 +9,7 @@ cat > "${tempFile}" <<EOF
         add_header X-Frame-Options "SAMEORIGIN";
         add_header X-Content-Type-Options "nosniff";
         add_header Referrer-Policy "strict-origin-when-cross-origin";
-        add_header Content-Security-Policy "default-src 'self'; style-src 'self' 'unsafe-inline'; connect-src 'self' blob:;";
+        add_header Content-Security-Policy "default-src 'self'; style-src 'self' 'unsafe-inline'; connect-src 'self' http://${API_HOST} https://${API_HOST} ws://${API_HOST} wss://${API_HOST} blob:;";
 EOF
 
 if [ "${ENABLE_SECURITY_HEADERS}" = "true" ]; then


packages/web/scripts/security.sh
```diff
@@ -0,0 +1,20 @@
...
+
+        add_header X-Frame-Options "SAMEORIGIN";
+        add_header X-Content-Type-Options "nosniff";
+        add_header Referrer-Policy "strict-origin-when-cross-origin";
+        add_header Content-Security-Policy "default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; frame-ancestors 'self'; connect-src 'self' http://${API_DOMAIN} https://${API_DOMAIN} ws://${API_DOMAIN} wss://${API_DOMAIN} blob:;";
+EOF
+
+if [ "${ENABLE_SECURITY_HEADERS}" = "true" ]; then
+  if grep -q "#SecurityHeaders" /etc/nginx/nginx.conf.tmpl; then
+    sed -i "/#SecurityHeaders/r ${tempFile}" /etc/nginx/nginx.conf.tmpl
```
rangoo94, line 14: Is it correct?
* There is missing closing `/` for sed expression
* There is file name prepended by `r ` provided instead of actuall contents

packages/web/scripts/security.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+tempFile=$(mktemp)
```
rangoo94, line 3: We shouldn't create new files outside of `/etc/nginx` - the rest of filesystem may be read-only

packages/web/scripts/security.sh
```diff
@@ -0,0 +1,20 @@
...
+cat > "${tempFile}" <<EOF
+
+        add_header X-Frame-Options "SAMEORIGIN";
+        add_header X-Content-Type-Options "nosniff";
+        add_header Referrer-Policy "strict-origin-when-cross-origin";
+        add_header Content-Security-Policy "default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; frame-ancestors 'self'; connect-src 'self' http://${API_DOMAIN} https://${API_DOMAIN} ws://${API_DOMAIN} wss://${API_DOMAIN} blob:;";
```
rangoo94, line 9: `script-src` and `frame-ancestors` are redundant to `default-src`

rangoo94: what is the `API_DOMAIN`? I don't think we are passing that there

packages/web/scripts/security.sh
```diff
@@ -0,0 +1,20 @@
...
+EOF
+
+if [ "${ENABLE_SECURITY_HEADERS}" = "true" ]; then
+  if grep -q "#SecurityHeaders" /etc/nginx/nginx.conf.tmpl; then
```
rangoo94, line 13: unnecessary `grep` check - `sed` will just do nothing if it will not find

packages/web/scripts/security.sh
devcatalin: The `sed` command does not require a closing / for the search pattern when using the r (read from file) command. The use of `r` followed by a space and a filename inserts the content of the file after the line where the pattern matches.

Tested multiple times and it works right..

packages/web/scripts/security.sh
devcatalin: I'll remove it for now, it's only needed if they deploy the API on a different domain than the dashboard.
It's harder to parse the hostname from REACT_APP_API_SERVER_ENDPOINT instead of adding a new API_DOMAIN url, but I don't want to repeat the logic either.. will see what their input is

